PR #6, opened 2025-01-26
Added modelence-setup CLI command 

rajiv-pat:
Added modelence-setup CLI command, also added MODELENCE_SERVICE_TELEMETRY environment variable to have a way enable/disable telemetry.

### DIFF
--- a/app/logs.ts
+++ b/app/logs.ts
@@ -1,7 +1,7 @@
 import { getLogger } from './metrics';
 
 export function logInfo(message: string, args: object) {
-  const isTelemetryEnabled = Boolean(process.env.MODELENCE_TELEMETRY_ENABLED);
+  const isTelemetryEnabled = process.env.MODELENCE_TELEMETRY_ENABLED === 'true';
 
   if (isTelemetryEnabled) {
     getLogger().info(message, args);
@@ -9,7 +9,7 @@ export function logInfo(message: string, args: object) {
 }
 
 export function logError(message: string, args: object) {
-  const isTelemetryEnabled = Boolean(process.env.MODELENCE_TELEMETRY_ENABLED);
+  const isTelemetryEnabled = process.env.MODELENCE_TELEMETRY_ENABLED === 'true';
 
   if (isTelemetryEnabled) {
     getLogger().error(message, args);
--- a/app/index.ts
+++ b/app/index.ts
@@ -30,7 +30,7 @@ export async function startApp(
   dotenv.config({ path: '.modelence.env' });
 
   const hasRemoteBackend = Boolean(process.env.MODELENCE_SERVICE_ENDPOINT);
-  const isCronEnabled = Boolean(Number(process.env.MODELENCE_CRON_ENABLED));
+  const isCronEnabled = process.env.MODELENCE_CRON_ENABLED === 'true';
 
   // TODO: verify that user modules don't start with `_system.` prefix
   const systemModules = [userModule, sessionModule, cronModule];
--- a/app/metrics.ts
+++ b/app/metrics.ts
@@ -16,7 +16,7 @@ export const initMetrics = async () => {
 
   isInitialized = true;
 
-  const isTelemetryEnabled = Boolean(process.env.MODELENCE_TELEMETRY_ENABLED);
+  const isTelemetryEnabled = process.env.MODELENCE_TELEMETRY_ENABLED === 'true';
   if (isTelemetryEnabled) {
     await initElasticApm();
   }
@@ -79,7 +79,7 @@ async function initElasticApm() {
 }
 
 export function startTransaction(type: 'method' | 'cron', name: string, context?: Record<string, any>) {
-  const isTelemetryEnabled = Boolean(process.env.MODELENCE_TELEMETRY_ENABLED);
+  const isTelemetryEnabled = process.env.MODELENCE_TELEMETRY_ENABLED === 'true';
 
   if (!isTelemetryEnabled) {
     return {


app/logs.ts
```diff
@@ -1,15 +1,15 @@
 import { getLogger } from './metrics';
 
 export function logInfo(message: string, args: object) {
-  const isTelemetryEnabled = Boolean(process.env.MODELENCE_SERVICE_ENDPOINT);
+  const isTelemetryEnabled = Boolean(process.env.MODELENCE_SERVICE_TELEMETRY);
```
artahian: Can we name this `MODELENCE_TELEMETRY_ENABLED` instead? The "Service" part is probably not needed because it was describing what kind of endpoint it is, but "telemetry" just on its own is pretty clear.

cli/commands/setup.ts
```diff
@@ -0,0 +1,43 @@
...
+export async function setupCommand(args: string[]) {
+  const tokenIndex = args.indexOf('--token');
+  if (tokenIndex === -1 || !args[tokenIndex + 1]) {
+    console.error('Error: --token parameter is required');
+    process.exit(1);
+  }
+
+  const token = args[tokenIndex + 1];
+  await updateEnvFile(token);
+}
```
artahian: Can we use something like https://github.com/tj/commander.js ? I asked Cursor what NPM libraries do this kind of thing and it recommended that one. Right now this is very simple but doing with commander should be more reliable/scalable and handle all corner cases without manually parsing.

cli/commands/setup.ts
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    const envPath = path.join(process.cwd(), '.env');
+    let envContent = '';
+    
+    try {
+      envContent = await fs.readFile(envPath, 'utf8');
+    } catch (error) {
+      // If file doesn't exist, start with empty content
+      envContent = '';
+    }
```
artahian: Can we use `dotenv` to parse the existing file?

https://www.npmjs.com/package/dotenv

We already use this to load the file to be used, looks like it has a `dotenv.parse` command that can just return the values. This way it will handle all cases with formatting, like extra whitespace, different quotes, etc

cli/commands/setup.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    // Check if MODELENCE_SERVICE_TOKEN already exists
+    if (envContent.includes('MODELENCE_SERVICE_TOKEN=')) {
```
artahian: We actually need the other values returned by the setup API too:

serviceEndpoint -> MODELENCE_SERVICE_ENDPOINT
containerId -> MODELENCE_CONTAINER_ID

and also just need to hardcode `MODELENCE_CRON_INSTANCE=1` to enable cron on dev environments.

cli/commands/setup.ts
```diff
@@ -0,0 +1,43 @@
...
+      envContent = envContent.replace(
+        /MODELENCE_SERVICE_TOKEN=["']?[^"'\n]*["']?/,
+        `MODELENCE_SERVICE_TOKEN="${token}"`
+      );
+    } else {
+      envContent += `${envContent ? '\n' : ''}MODELENCE_SERVICE_TOKEN="${token}"`;
```
artahian: I'm thinking about how we can adjust the file without breaking other existing things, like formatting, comments etc, since there can be pretty much anything in the existing `.env` file.

Here's an idea that I've seen some tools use - sometimes they create a section like this:
```
# ===== BEGIN MODELENCE SETUP =====
# WARNING: This section is automatically managed. Do not edit manually.
# Changes will be overwritten.

[content goes here]

# ===== END MODELENCE SETUP =====
```



[EDIT]: I just had a better idea - can we just use a separate `.modelence.env` file and be done with it? This would be so much easier. We can adjust the usage to be like this:

```
dotenv.config();

dotenv.config({ path: '.modelence.env' });
```

to load both .env and .modelence.env

If `.modelence.env` already exists when running the setup CLI, it can show a warning to the user and ask them to confirm if they want to overwrite the existing file.  This way we would not have to do any search/replace, would be very simple override each time.

package.json
```diff
@@ -8,11 +8,16 @@
...
+    "./server": "./dist/server.js",
+    "./cli": "./dist/cli/index.js"
   },
   "files": [
-    "dist"
+    "dist",
+    "dist/cli"
   ],
+  "bin": {
+    "modelence-setup": "./dist/cli/index.js"
```
artahian: Also, can we name this `setup.js` instead of `index.js`? We might have other scripts later, so it will be easier to keep track that way.

cli/commands/setup.ts
rajiv-pat: I think MODELENCE_SERVICE_ENDPOINT is going to be fixed at some point right? It should be hardocded. to cloud.modelence.com. Should we just hardcode that value?
MODELENCE_CONTAINER_ID seems useless to me, or I don't get the meaning of it fully. Shouldn't the token be enough?

artahian: I think at least it's still good to have in env for running studio locally (or connecting to a staging environment, etc, that might have a different endpoint). My original intention was that maybe at some point it will be distributed among multiple regions like many similar services and have a region-specific endpoint (like us-east-1.cloud.modelence.com or something similar).

cli/setup.ts
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
+import { Command } from 'commander';
+import { setupCommand } from './commands/setup';
```
artahian: Just curious - why do we need two setup.ts files? Can everything just be in the cli/setup.ts file itself?

cli/commands/setup.ts
```diff
@@ -0,0 +1,66 @@
...
+    rl.question(`Warning: ${MODELENCE_ENV_FILE} already exists. Do you want to overwrite it? (y/N) `, (answer) => {
+      rl.close();
+      resolve(answer.toLowerCase() === 'y');
+    });
+  });
+}
+
+export const setupCommand = new Command('setup')
+  .description('Setup Modelence environment variables')
+  .requiredOption('-t, --token <token>', 'Modelence service token')
```
artahian: Looks like this should be the setup token which is different from the service token.
```suggestion
  .requiredOption('-t, --token <token>', 'Modelence setup token')
```

cli/commands/setup.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      try {
+        // Check if .modelence.env exists
+        const envContent = await fs.readFile(envPath, 'utf8');
+        existingEnv = dotenv.parse(envContent);
+
+        // Ask for confirmation before overwriting
+        const shouldContinue = await confirmOverwrite();
+        if (!shouldContinue) {
+          console.log('Setup cancelled');
```
artahian: Slight adjustment for American vs British version
```suggestion
          console.log('Setup canceled');
```

cli/commands/setup.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+      } catch (error) {
+        // File doesn't exist, we'll create it
+      }
+
+      // Update or add the variables
+      const newEnv = {
+        ...existingEnv,
+        MODELENCE_CRON_INSTANCE: 1,
+        MODELENCE_SERVICE_TOKEN: options.token
```
artahian: This is actually the setup token, not the service token, we still need to make an API call with a POST request to Studio's `/api/setup` endpoint with a `X-Modelence-Setup-Token` header set to the value of the setup token, and this will API will return the service token along with the other values like endpoint and container ID.

cli/commands/setup.ts
```diff
@@ -0,0 +1,66 @@
...
+      };
+
+      // Convert to .env format
+      const envContent = Object.entries(newEnv)
+        .map(([key, value]) => `${key}="${value}"`)
```
artahian: Probably won't be an issue in most cases but I think we need to escape all `"` characters in value otherwise it will break if it contains quotes.

cli/commands/setup.ts
rajiv-pat: ooh, now I got what you mean. All makes sense now to me. Will change it to that API call.

cli/setup.ts
```diff
@@ -0,0 +1,107 @@
...
+interface SetupResponse {
+  deploymentId: string;
+  serviceEndpoint: string;
+  serviceToken: string;
+  containerId: string;
+}
+
+async function fetchServiceConfig(setupToken: string, endpoint: string): Promise<SetupResponse> {
+  // Using global fetch (available in Node.js 18+)
+  const response = await globalThis.fetch(`${endpoint}/api/setup`, {
```
artahian: I believe this is the same as just `fetch` without `globalThis`?

cli/setup.ts
```diff
@@ -0,0 +1,107 @@
...
+function escapeEnvValue(value: string | number): string {
+  // Convert to string and escape quotes
+  return String(value).replace(/"/g, '\\"');
+}
+
+const program = new Command()
+  .name('modelence-setup')
+  .description('Setup Modelence environment variables')
+  .requiredOption('-t, --token <token>', 'Modelence setup token')
+  .option('-e, --endpoint <endpoint>', 'Modelence endpoint', 'cloud.modelence.com')
```
artahian: Since this is not the full endpoint, maybe it is more correct to say "host" or "base-url" instead?

cli/setup.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      // Fetch service configuration using setup token
+      console.log('Fetching service configuration...');
+      const config = await fetchServiceConfig(options.token, options.endpoint);
+
+      // Update environment variables
+      const newEnv = {
+        ...existingEnv,
+        MODELENCE_CRON_INSTANCE: 1,
+        MODELENCE_TELEMETRY_ENABLED: '',
```
artahian: can we make this in the same format as `MODELENCE_CRON_INSTANCE`?  I'm not sure what's the standard for environment variables, but looks like just 0 and 1 is not a good option either because they are usually interpreted as string and '0' is going to turn into `true`.

cli/setup.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Using global fetch (available in Node.js 18+)
+  const response = await globalThis.fetch(`${endpoint}/api/setup`, {
+    method: 'GET',
+    headers: {
+      'X-Modelence-Setup-Token': setupToken,
+    },
+  });
+
+  if (!response.ok) {
+    throw new Error(`Setup failed: ${response.statusText}`);
```
artahian: Also looks like we don't need this `Setup failed` prefix here also, otherwise I got something like this:

```
Setup failed: Setup failed: Internal Server Error
```

artahian: And also, it just shows `Internal Server Error` now, we need to show the actual error message "Error: Setup token is invalid or expired", which is returned in `response.text()`.

cli/setup.ts
```diff
@@ -76,15 +76,15 @@ const program = new Command()
 
       // Fetch service configuration using setup token
       console.log('Fetching service configuration...');
-      const config = await fetchServiceConfig(options.token, options.endpoint);
+      const config = await fetchServiceConfig(options.token, options.host);
 
       // Update environment variables
       const newEnv = {
         ...existingEnv,
-        MODELENCE_CRON_INSTANCE: 1,
```
artahian: I see you're changing from `MODELENCE_CRON_INSTANCE` to `MODELENCE_CRON_ENABLED`, do we need to change the usage in the framework also? I think right now it's looking for `MODELENCE_CRON_INSTANCE`.

rajiv-pat: good catch!

PR #9, opened 2025-02-13
Refactor CLI and add deploy command

artahian:
I was adding a `deploy` command alongside `setup` and realized that there is a simple way to group all these CLIs together, similar to how Next.js does it. So I moved it all into a `bin` folder (since that seems to be the standard convention) and grouped all under a single `modelence` cli command, so now we can run `modelence setup` instead of `modelence-setup`, but also `modelence deploy` and other commands later by having just one entry point instead of having multiple CLI entry points.

For the deploy, it is currently just creating a zip file from the build without doing anything - later we will need to upload it to S3 and push to AWS ECR.

Also, I wanted to make the `.modelence` folder more general instead of using it only for the build, so I moved the build from there into a `.modelence/build` subdirectory. Now we can use `.modelence` for other things too - like storing the zipped bundle itself and things like user's current Modelence Studio session (for deploying authorization).

### DIFF
--- a/bin/deploy.ts
+++ b/bin/deploy.ts
@@ -142,6 +142,7 @@ async function triggerDeployment(bundleName: string) {
   const response = await fetch(getStudioUrl(`/api/deployments/${deploymentId}/deploy`), {
     method: 'POST',
     headers: {
+      'Content-Type': 'application/json',
     },
     body: JSON.stringify({
       bundleName,


bin/setup.ts
```diff
@@ -0,0 +1,111 @@
+import { promises as fs } from 'fs';
```
artahian, line 1: I moved this file without changes (except from removing the CLI part which is now handled in the main modelence.ts script).

tsup.config.ts
```diff
@@ -1,15 +1,14 @@
...
-  entry: ['index.ts', 'client.ts', 'server.ts', 'cli/setup.ts'],
+  entry: ['index.ts', 'client.ts', 'server.ts', 'bin/modelence.ts'],
   format: ['esm'],
   dts: {
     resolve: true,
     entry: {
       index: 'index.ts',
       client: 'client.ts',
-      server: 'server.ts',
-      cli: 'cli/setup.ts'
```
artahian, line 12: I removed this altogether, not sure if it's needed since this was for exporting TypeScript types which shouldn't apply to CLI scripts.

bin/deploy.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  await archive.finalize();
+  await archiveComplete;
+
+  const stats = await fs.stat(outputFile);
+  console.log(`Deployment bundle created at: ${outputFile} (${(stats.size / 1024 / 1024).toFixed(2)} MB)`);
+
+  // TODO: Add authentication
+  const deploymentId = '677d93fe2cdf304863f3a0f6';
+  const response = await fetch(`http://localhost:3000/api/deployments/${deploymentId}/deploy`, {
```
rajiv-pat: Can't we use MODELENCE_SERVICE_ENDPOINT env variable here directly? As it should be defined already in the .modelence.env file and we should use the same endpoint here, right?

artahian: This part is still a draft and I was going to make more changes (at least to add authentication).

But I'm not sure if we have access to that endpoint, because the `.env` file is auto-loaded into the application itself, but this is a separate script running directly, triggered from the terminal as a CLI.

I'll take a look - I'm wondering if we should move some or all of these values from `.env` into a `modelence.config.js` file instead which will be easier to read. But in the worst case we can still manually load and use `.modelence.env` from the script as well.

PR #10, opened 2025-02-17
CLI authentication flow for `modelence deploy`

artahian:
This PR adds a device authorization flow and CLI token authentication for both build upload and deployment. The corresponding PR for Studio is here: https://github.com/modelence/studio/pull/8

I've refactored some of the existing code into a separate auth.ts and config.ts files. Now the script will always trigger the authentication flow by generating a temporary code, then redirecting the user to login and verify that code in the browser. After that the script keeps polling to see when a token is associated with the code. It then stores the token in `.modelence/auth.json` (in the future we can reuse this and not prompt to go through this flow every time the user deploys).

### DIFF
--- a/app/metrics.ts
+++ b/app/metrics.ts
@@ -113,6 +113,11 @@ export function startTransaction(type: 'method' | 'cron', name: string, context?
 }
 
 export function captureError(error: Error) {
+  if (!isTelemetryEnabled()) {
+    console.error(error);
+    return;
+  }
+
   if (!apm) {
     throw new Error('captureError: Elastic APM is not initialized');
   }


bin/deploy.ts
```diff
@@ -2,9 +2,8 @@ import { createWriteStream, promises as fs } from 'fs';
...
 import { join } from 'path';
 import archiver from 'archiver';
-import { parse as parseDotenv } from 'dotenv';
-
-let studioBaseUrl = '';
+import { authenticateCli } from './auth';
+import { getStudioUrl } from './config';
 
 // TODO: Determine dynamically
 const deploymentId = '677d93fe2cdf304863f3a0f6';
```
rajiv-pat: Can't we change this to MODELENCE_DEPLOYMENT_ID which should be already available in the .modelence.env file?
Every time I need to change this hardcoded value to my deployment_id :)

artahian: Ah right I forgot about this one - but no we can't really use `MODELENCE_DEPLOYMENT_ID` because it was a temporary hack to just make the dashboard work for a single hardcoded application. I'll think about this a little more and adjust before we merge, the challenge is that deploymentId is not really available locally. It can be derived from the Modelence service token, but there is more complexity here which we probably should talk about:

- When the user triggers the deployment, they are in their local environment, using a .modelence.env configuration tied to a local dev deployment.
- The deployment process is intended to deploy production, which is a different environment (there is no such thing as "deploy" for local dev environments in Studio).
- Question: how does this actually work then? I think we should also somehow specify which environment we're deploying to, because there can be more than one cloud environment for the same app (production, staging, demo, etc). Maybe we can use the `deploymentAlias` and require it, so it becomes something like `modelence deploy production` (given that `production` is the deployment alias for production). The application itself can be automatically determined by the Modelence service token, but this means that it won't work for local dev environments that are not connected to Studio - maybe I just want to have an offline dev setup without Studio, but still deploy it to production.

artahian: To address this last part, we could just pass everything explicitly, but it requires passing both every time and I wonder if users might accidentally mix up the command and deploy the wrong application with some other project's code. Meteor had the same problem though, so this is probably not an absolute dealbreaker:

```sh
modelence deploy typesonic production
```

or 

```sh
modelence deploy --app typesonic --env production
```

PR #17, opened 2025-04-17
Add app start tracking

artahian:
Added a tracking call on app startup to understand the framework usage and estimate how many users are using Modelence and which version.

The tracking can be disabled by setting `MODELENCE_TRACKING_ENABLED` to `false` in environment variables.

### DIFF
--- a/packages/modelence/src/app/index.ts
+++ b/packages/modelence/src/app/index.ts
@@ -39,6 +39,12 @@ export async function startApp(
   const hasRemoteBackend = Boolean(process.env.MODELENCE_SERVICE_ENDPOINT);
   const isCronEnabled = process.env.MODELENCE_CRON_ENABLED === 'true';
 
+  trackAppStart().then(() => {
+    // Do nothing
+  }).catch(() => {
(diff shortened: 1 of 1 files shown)


packages/modelence/src/app/index.ts
```diff
@@ -175,3 +181,23 @@ function getLocalConfigs(): AppConfig[] {
 
   return configs;
 }
+
+async function trackAppStart() {
+  const isTrackingEnabled = process.env.MODELENCE_TRACKING_ENABLED !== 'false';
+
+  if (isTrackingEnabled) {
+    const serviceEndpoint = process.env.MODELENCE_SERVICE_ENDPOINT ?? 'https://cloud.modelence.com';
```
rajiv-pat, line 189: If this is only for statistics, isn't it better to always send that data to cloud.modelence.com no matter if they have a service endpoint or not? I think in 99% of the cases it will work, but if in the future we will introduce something like cloud-demo and share it with users, it will split part of the data to demo and part to cloud.

hello-world2025 (approved):
Added one small comment, but approving.

PR #21, opened 2025-06-06
AI SDK

artahian:
- Added @modelence/ai as a new package (more details in its own README.md). Ended up wrapping Vercel AI SDK so we can get full control around telemetry and tracking. Already published on NPM (v0.1.1)
- Moved modelence into peer dependencies to fix problems with having multiple versions of modelence (for both `ai` and `auth-ui` packages).
- Added a `CloudBackendConnectResponse` type used by Studio to avoid type mismatches (turned out we still used the old `deploymentId` and `deploymentAlias` instead of `environmentId` and `environmentAlias`.
- Moved some telemetry code in modelence into a separate telemetry.ts file and exposed via separate `modelence/telemetry` import. This was needed for `@modelence/ai` to do tracing.


hello-world2025 (approved):
All looks good, and changes are pretty simple!

PR #22, opened 2025-06-18
Mintlify documentation

artahian:
Ok, I don't even know where to start... basically I replaced Docusaurus with Mintlify and somehow made the Typedoc generation fully work to create a complete API Reference from all our classes, functions, types, etc.

The changes look very big, but it's mostly just removing Docusaurus files and adding Mintlify's starter code.

Check out the working version of docs here: https://modelence.mintlify.app/

It is configured now serve from the `_docs` branch, which is auto-generated on every push on `main`.


rajiv-pat (approved):
All looks good to me. And I approve this.
But when reviewing the documentation, I have a feeling that pure API references are not that helpful. Those that have examples are very helpful, e.g., https://modelence.mintlify.app/api-reference/@modelence/react-query/functions/modelenceMutation but some contain only pure type/method definition, and look a little strange to me https://modelence.mintlify.app/api-reference/modelence/client/functions/callMethod

artahian:
Yeah I agree that the ones without examples are not very useful, the point for now is to just have some starting documentation that we'll improve later. This will also be helpful for LLM bots that will be scraping the docs.

PR #23, opened 2025-06-19
Update package.json

omegascorp:
Remove the dependency of the typescript from `npx create-modelence-app`


artahian:
@omegascorp what exactly does the `--bundle` flag do? I'm not sure I fully understand why there was an issue with TypeScript before (given that it's in dev dependencies).

omegascorp:
@artahian I'm closing this it's not relevant

PR #25, opened 2025-06-22
Block disposable email addresses during signup

artahian:
Prevents users from signing up with throwaway email addresses by checking against a database of ~100k known disposable domains.

## What it does
- Automatically fetches and updates disposable email domains daily from [disposable.github.io](https://disposable.github.io/disposable-email-domains/domains.txt)
- Blocks signup attempts using domains like `tempmail.com`, `10minutemail.com`, `guerrillamail.com`, etc.
- Shows users a clear error: *"Please use a permanent email address"*

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.5.2",
+  "version": "0.5.3-dev.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.5.2",
+      "version": "0.5.3-dev.0",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@octokit/rest": "^20.0.2",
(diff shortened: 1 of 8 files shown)


artahian:
@omegascorp 

> I was previously checking databases of disposable email domains; for example, this one https://github.com/disposable-email-domains/disposable-email-domains/blob/main/disposable_email_blocklist.conf contains approximately 4,300 domains.
> 
> The issue is that half of the disposable domains I was getting on Alienart.io were not in the list. So we will need to extend it over time.

The database we're using in this PR has 108,007 domains, and it is updated on an almost daily basis, so I believe it should have a much better coverage. If you know any disposable domains that weren't caught we can check against this full list, but this seems to be the most comprehensive.

> Another issue with this approach is that once the database is created, we will no longer be able to add new domains to the client's db.

We're not doing this just one time:
> Automatically fetches and updates disposable email domains daily from [disposable.github.io](https://disposable.github.io/disposable-email-domains/domains.txt)

there is a cron job that keeps the list updated daily. When the user first runs their app, the cron job runs immediately and initializes the list, but it also keeps it updated (it's a system cron job that the user doesn't even see).

omegascorp (approved):
Ok, looks good

PR #27, opened 2025-06-25
Auth rate limits

### LINKED ISSUE #29: Implement rate limiting
For signups, but also generally available for any usage by Modelence apps

artahian:
Closes #29 

This PR adds a general rate limit support for Modelence, and also implements it for user signups. The interface is pretty simple - you just need to pass a `rateLimits` array to a Module, like this:
```js

new Module('_system.user', {
  ...
  rateLimits: [{
    bucket: 'signup',
    type: 'ip',
    window: time.minutes(15),
    limit: 20,
  }, {
    bucket: 'signup',
    type: 'ip',
    window: time.days(1),
    limit: 200,
  }],
});
```

And then later check it using `consumeRateLimit`:
```js
import { consumeRateLimit } from 'modelence/server';
...
await consumeRateLimit({
  bucket: 'signup',
  type: 'ip',
  value: ip,
});
```

### Implementation
I've ended up using a two-window approach. We determine the window block by rounding the time of the event (i.e. if the window is 5m, we get x:00 - x:05, x:05 - x:10 blocks, etc). We only store the counts of events for the last 2 blocks - the block that the last event falls into, and the one before that. At any given time, we can approximate the number of events that happened between [now - windowSize, now]. Here's an example:
- Window size: 5 mins
- [10:10 - 10:15] - 3 events, [10:15 - 10:20], 2 events.
- Current time: 10:19 (so we're looking to estimate events that happened between [10:14 - 10:19]
- All events in the last block are included since they are fully contained
- To approximate the number of events from the previous block, we use a weighted average. [10:14-10:19] has a 1 minute overlap with the previous block, which comes to a 1/5 = 0.2 weight. So estimated number of events from the previous block will be 3 * 0.2 = 0.6.
- So, the total is 2 + 0.6 = 2.6 events
- Finally, we round this to get 3

This assumes events in each block are evenly spaced, so again, this is heuristic approximation, but it is a standard rate limit calculation method and is ok because the goal is to avoid abuse, even if we're approximating a little.

Also, we're currently just storing this in a MongoDB collection. In the future, once we need to support a much larger scale, we should consider moving this into Redis or another similar storage that's faster and a better fit for this. But, I believe the current implementation will be very solid even for significant scale for a while.

### Testing

To test locally, I added a smaller rate limit for convenience:
```
    bucket: 'signup',
    type: 'ip',
    window: time.minutes(5),
    limit: 2,
```

---

Follow-ups deferred to later: https://github.com/modelence/modelence/issues/31

### DIFF
--- a/packages/modelence/src/server.ts
+++ b/packages/modelence/src/server.ts
@@ -18,6 +18,7 @@ export { CronJobInputParams } from './cron/types';
 
 // Rate limits
 export type { RateLimitRule } from './rate-limit/types';
+export { consumeRateLimit } from './rate-limit/rules';
 
 export { getConfig } from './config/server';
 export type { CloudBackendConnectResponse } from './app/backendApi';


omegascorp:
Just as a note cloudflare has rate-limit and captcha in the free version https://developers.cloudflare.com/waf/rate-limiting-rules/

packages/modelence/src/rate-limit/db.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    windowStart: schema.date(),
+    windowCount: schema.number(),
+    prevWindowCount: schema.number(),
+
+    expiresAt: schema.date(),
+  },
+  indexes: [
+    { key: { bucket: 1, type: 1, value: 1, windowMs: 1 }, unique: true },
+    { key: { expiresAt: 1 }, expireAfterSeconds: 0 },
```
artahian, line 25: This helps keep the collection clean. Once the latest captured window / count goes beyond the current timeline, it will be cleaned up.

artahian:
> Just as a note cloudflare has rate-limit and captcha in the free version https://developers.cloudflare.com/waf/rate-limiting-rules/

@omegascorp makes sense, but I think this is too rigid for us to use. By having rate limits in the framework itself we get full control on any type of an action + we can offer the same functionality to framework users to define their own rate limits.
With Cloudflare, it is almost in the infra layer without much context about the application itself and types of actions, and we wouldn't be able to customize it or make it configurable via Modelence Cloud.

PR #30, opened 2025-06-25
Google Sign In

omegascorp:
* added google oauth routes
* it uses local config (it was a fast solution, we can add config to the UI in a separate PR)
* now the app reads auth token both from cookies and the header


packages/modelence/src/auth/providers/google.ts
```diff
@@ -0,0 +1,103 @@
...
+  res.status(301);
+  res.redirect("/");
+}
+
+
+async function handleGoogleAuthenticationCallback(req: Request, res: Response) {
+  const googleUser = req.user as GoogleUser;
+
+  const existingUser = await usersCollection.findOne(
+    { 'google.id': googleUser.id },
```
artahian: I think this should be `authMethods.google.id`?

packages/modelence/src/auth/providers/google.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  const existingUserByEmail = await usersCollection.findOne(
+    { 'emails.address': googleUser.email, },
+    { collation: { locale: 'en', strength: 2 } },
+  );
+
+  // TODO: check if the email is verified
+  if (existingUserByEmail) {
+    // If the user exists but has a different Google ID, we can link the Google account
```
artahian: Can we just throw an error in this case? We used to do it that way at CodeSignal, because we don't know for sure if it's the same user. Also, I may have somehow connected my X@gmail.com google account to a user account with my Y@gmail.com email address, so we'll be overriding it.

I believe the right way to link is to explicitly prompt the user to link their existing account:
- User logs in via Google
- We detect the email is already registered, ask them if they want to merge
- If they say yes, they log in with email/password as the next step to their existing account
- Google account is not linked to the existing account

But since this proper flow is a little complicated we can just omit it for now and throw an error instead, saying sign up with password instead.

packages/modelence/src/auth/providers/google.ts
```diff
@@ -0,0 +1,103 @@
...
+      { _id: existingUserByEmail._id },
+      { $set: { google: { id: googleUser.id } } }
+    );
+    await authenticateUser(res, existingUserByEmail._id);
+    return;
+  }
+
+  // If the user does not exist, create a new user
+  const newUser = await usersCollection.insertOne({
+    handle: googleUser.displayName,
```
artahian: What if the handle is taken? We probably need some extra checks here?

packages/modelence/src/auth/providers/google.ts
```diff
@@ -0,0 +1,103 @@
...
+    { 'emails.address': googleUser.email, },
+    { collation: { locale: 'en', strength: 2 } },
+  );
+
+  // TODO: check if the email is verified
+  if (existingUserByEmail) {
+    // If the user exists but has a different Google ID, we can link the Google account
+    await usersCollection.updateOne(
+      { _id: existingUserByEmail._id },
+      { $set: { google: { id: googleUser.id } } }
```
artahian: Similar to the previous comment, shouldn't this be `authMethods.google`?

packages/auth-ui/src/components/LoginForm.tsx
```diff
@@ -64,6 +67,7 @@ export function LoginForm({
             variant="outline" 
             className="w-full flex items-center justify-center gap-3"
             type="button"
+            onClick={openGoogleLogin}
           >
             <GoogleIcon className="w-5 h-5" />
             <span className="font-medium">Sign in with Google</span>
```
artahian: Should we add this to the SignupForm as well? I think usually social logins are offered as an option in both login & signup.

artahian: We need to also make these buttons optional and fetch some info from the server from the social auth configuration. If Google tokens / keys are not configured, it should not show this button as an option, because otherwise most apps will have a broken google button unless they add the keys.

artahian (changes requested):
Aside from the comments, a more high level question - could we potentially use https://www.better-auth.com or is it not customizable enough for us to build on?

packages/modelence/src/app/index.ts
```diff
@@ -145,6 +145,10 @@ function getRateLimits(modules: Module[]) {
 function getConfigSchema(modules: Module[]): ConfigSchema {
   const merged: ConfigSchema = {};
 
+  for (const [key, value] of Object.entries(systemConfigSchema)) {
+    merged[key] = value;
+  }
+
```
artahian: `modules` passed to this function already include system modules, so this doesn't seem right - I think we should derive system config schema the same way. The current problem is that system modules are simply not defining their own config schema as needed.

packages/modelence/src/app/systemConfigSchema.ts
```diff
@@ -0,0 +1,21 @@
+import { ConfigSchema } from "@/config/types";
```
artahian: Related to the other comment: I don't think we should put all configs here in one place. We already use a modular system, so each module is supposed to define its own config schema. The current problem is that system modules use some configs but they don't seem to properly define them. So the right way to do this would be to either use an existing module (like _system.user) or create a new _system.auth module and pass these into it as a config schema.

artahian: Example from Typesonic:

```ts
const configSchema: ConfigSchema = {
  'format.maxLineLength': {
    type: 'number',
    default: 40,
    isPublic: true,
  },
};

export default new Module('typewriterText', {
  stores: [dbTexts],
  queries,
  cronJobs,
  configSchema
});

```

packages/modelence/src/app/index.ts
```diff
@@ -154,6 +158,10 @@ function getConfigSchema(modules: Module[]): ConfigSchema {
         );
       }
 
+      if (key.toLowerCase().startsWith('_system.')) {
+        throw new Error(`Config key cannot start with a reserved prefix: '_system.' (${key})`);
+      }
+
```
artahian: I think this won't work because we already namespace configs with the corresponding module name. So currently if you have a module called `_system.auth`, it would just need to pass `google.apiKey` as the config name (see the `absoluteKey` calculation above).

packages/modelence/src/app/index.ts
```diff
@@ -180,15 +188,26 @@ function initStores(stores: Store<any, any>[]) {
...
 
-  if (process.env.MONGODB_URI) {
-    configs.push({
-      key: '_system.mongodbUri',
-      type: 'string',
-      value: process.env.MONGODB_URI
-    });
+  for (const [envVar, configKey] of Object.entries(localConfigMap)) {
+    const value = process.env[envVar];
+    console.log(configKey, value);
```
artahian: Leftover?

artahian (changes requested):

packages/modelence/src/app/systemConfigSchema.ts
omegascorp: Ok, I'm moving the config to user module

packages/modelence/src/app/index.ts
omegascorp: Yeah, removed from both places. It's really not required

artahian (approved):

PR #32, opened 2025-06-26
Temporary hide extra links

omegascorp:
* Hidden Google auth, reset password
* Removed m@example.com (I've checked other websites in most cases, when there is an Email label, they don't give an example of email)


artahian (commented):
@omegascorp I actually made these same changes already in https://github.com/modelence/modelence/pull/33 in a slightly different way. But for the broken links I actually meant the ones here: https://github.com/modelence/modelence/blob/main/packages/auth-ui/src/components/UserProfile.tsx

I think the only change here that the other branch doesn't already contain is the m@example.com placeholder, so I will just manually apply that there.

PR #35, opened 2025-06-30
Add modelence config

omegascorp:
There is an issue with the configs shared between Modelence and its other packages. At this point, we have handleError, which is set on a local property. When, for some reason, versions do not match, we get multiple versions of the configs, which breaks some of the functionality.

The issue we currently face is that when we link auth-ui to a local version, it installs its own modelence, resulting in two copies of modelence.

### DIFF
--- a/packages/modelence/src/client/config.ts
+++ b/packages/modelence/src/client/config.ts
@@ -1,0 +1,37 @@
+declare global {
+  var modelenceConfig: ModelenceConfig;
+}
+
+function getGlobalObject() {
+  if (typeof global !== 'undefined') {
+    return global;
+  }
+  if (typeof window !== 'undefined') {
+    return window;
+  }
+  if (typeof self !== 'undefined') {
+    return self;
+  }
+  return {
+    modelenceConfig: {} as ModelenceConfig,
+  };
+}
+
+const globalObject = getGlobalObject();
+
+globalObject.modelenceConfig = {};
+
+interface ModelenceConfig {
+  errorHandler?: (error: Error, methodName: string) => void
+}
+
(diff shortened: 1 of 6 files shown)


artahian (commented):
Sounds great, but does this mean we have to avoid using local declarations everywhere? That seems to only fix this one case but there are a lot of other places. Shouldn't the multi-version configuration be impossible instead of us trying to completely avoid using local config everywhere, which will be almost impossible to enforce?

omegascorp:
Not necessarily, so far it's the only side-effect I've noticed, if we notice it in other places we will just need to add it to this config file

packages/modelence/src/client/config.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  if (typeof window !== 'undefined') {
+    return window;
+  }
+  if (typeof self !== 'undefined') {
+    return self;
+  }
+  return {
+    modelenceConfig: {} as ModelenceConfig,
+  };
```
artahian, line 17: Shouldn't we just throw an error in this case? If we're returning a new empty object each time, this is probably not going to work as expected anyway. And this is client only code, so this should never be the case.

packages/modelence/src/client/config.ts
```diff
@@ -0,0 +1,37 @@
+declare global {
+  var modelenceConfig: ModelenceConfig;
```
artahian, line 2: Can we call this something else? We already have configs and it does something completely different (i.e. app configs).

artahian (approved):
Approving, but let's please call this something else other than config because we already have "configs" that means something else and it will create confusion.

PR #37, opened 2025-07-10
Email Providers

omegascorp:
* Add Resend
* Add Amazon SES
* Add SMTP support


packages/modelence/package.json
```diff
@@ -79,7 +83,10 @@
     "zustand": "^5.0.2"
   },
   "peerDependencies": {
+    "@aws-sdk/client-ses": ">=3.0.0",
+    "nodemailer": ">=7.0.0",
     "react": ">=18.0.0",
-    "react-dom": ">=18.0.0"
+    "react-dom": ">=18.0.0",
+    "resend": ">=4.0.0"
```
artahian: @omegascorp I thought we were going to create separate packages for each of these (i.e. `@modelence/resend`, etc)?

If we make them peer dependencies it will force everyone to have every single one installed, even though it all should be optional.

packages/modelence/src/app/index.ts
```diff
@@ -184,6 +184,14 @@ const localConfigMap = {
...
   GOOGLE_AUTH_CLIENT_ID: '_system.user.auth.google.clientId',
   GOOGLE_AUTH_CLIENT_SECRET: '_system.user.auth.google.clientSecret',
+  RESEND_API_KEY: '_system.email.resend.apiKey',
+  SES_REGION: '_system.email.ses.region',
+  SES_ACCESS_KEY_ID: '_system.email.ses.accessKeyId',
+  SES_SECRET_ACCESS_KEY: '_system.email.ses.secretAccessKey',
+  SMTP_HOST: '_system.email.smtp.host',
+  SMTP_PORT: '_system.email.smtp.port',
+  SMTP_USER: '_system.email.smtp.user',
+  SMTP_PASS: '_system.email.smtp.pass',
```
artahian: Can we prefix these configs (at least the new ones) with `MODELENCE_`? And specifically `MODELENCE_EMAIL` to match the full config key? For example `MODELENCE_EMAIL_SMTP_HOST`.
Otherwise I'm worried this will collide with other environment variables that were not meant for Modelence.

packages/modelence/src/email/loadProvider.ts
```diff
@@ -0,0 +1,45 @@
+import { EmailProvider, EmailProviderName } from './types';
+
+export async function loadProvider(name: EmailProviderName): Promise<EmailProvider> {
```
artahian: Dynamically importing these based on an input string seems problematic, because we would have to include every single email provider in the bundle. Instead of doing this can we just create a separate npm package wrapper for each provider, that exports its own `sendEmail` method? We can define a unified interface that each of them have to implement, but they don't even have to exactly match in case some providers have core differences.

I remember we discussed this earlier and decided to go with that separate package approach to avoid bundling providers with the core framework.

packages/modelence/src/email/providers/AmazonSesProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+  SendEmailCommand,
+} from '@aws-sdk/client-ses';
+import { EmailData, EmailProvider } from '../types';
+import { getConfig } from '@/server';
+
+export class AmazonSesProvider implements EmailProvider {
+  private client: SESClient;
+
+  constructor() {
+    const region = String(getConfig('_system.email.ses.region')) || 'us-east-1';
```
artahian: Can we make this config required? I think defaulting to `us-east-1` can be confusing if someone forgets to specify the region and then is not able to figure out why it doesn't work (if they are running in another region).

packages/modelence/package.json
omegascorp: It's not forcing to have every package installed. If you choose a provider that is not installed, it will throw an exception with instructions on how to install the provider package. I think there is not much value in having separate packages for every provider: if we make them have a direct dependency, we will need to periodically update those packages and will create a routine load for ourselves. If we make them have peer dependencies, it's not much different from what is done now. Also, in this case, you can just switch a variable to change the provider. In case of multiple modules, you will need to update a bunch of imports.

packages/modelence/src/email/loadProvider.ts
omegascorp: Duplicating the text here, too, it does not require installing all packages. There are just 3-5 lines of code per package, I don't want to overcomplicate this.

packages/smtp/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+ * sendEmail({
+ *  from: 'test@example.com',
+ *  to: 'test@example.com',
+ *  subject: 'Test Email',
+ *  html: '<h1>Hello World</h1>'
+ * })
+ * ```
+ * 
+ * @param data - The email payload containing sender, recipient, subject, and HTML content.
+ * @returns Query options object for TanStack Query's useQuery
```
artahian: Is this a leftover?

omegascorp: Yes, the PR is not finalized yet, as I mentioned I need to make sure the docs are correct

docs/web/docs.json
```diff
@@ -38,7 +38,10 @@
               "api-reference/modelence/index",
               "api-reference/@modelence/react-query/index",
               "api-reference/@modelence/ai/index",
-              "api-reference/@modelence/next/index"
+              "api-reference/@modelence/next/index",
+              "api-reference/@modelence/smtp/index",
+              "api-reference/@modelence/resend/index",
+              "api-reference/@modelence/aws-ses/index"
```
artahian: Can we move these 3 into a separate group called `Email`?

packages/aws-ses/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Installation
+
+```bash
+npm i @modelence/aws-ses
+```
+
+## Overview
+
+This package provides `sendEmail` function that utilizes resend under the hood. The configuration can be set via Modelence Cloud or the following environment variables:
```
artahian:
```suggestion
This package provides `sendEmail` function that utilizes **resend** under the hood. The configuration can be set via Modelence Cloud or the following environment variables:
```

It says "utilizes resend"?

packages/aws-ses/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Advanced example
+
+```tsx
+import { sendEmail } from '@modelence/resend';
+
+sendEmail({
+  to: 'test@example.com',
+  from: 'test@example.com',
+  subject: 'Test Email',
+  react: <h1>Test Email</h1>,
```
artahian: Is it generally a standard practice to call this `react` instead of `body` (i.e. after the structure type)?

packages/resend/README.md
```diff
@@ -0,0 +1,48 @@
+# @modelence/resend
+
+Resend Adapter for modelence
```
artahian:
```suggestion
[Resend](https://resend.com/) Adapter for modelence
```

packages/resend/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export type EmailPayload = {
+  from: string;
+  to: string;
+  subject: string;
+  html?: string;
+  text?: string;
+  react?: ReactNode;
+  cc?: string;
+  bcc?: string;
```
artahian: What about multiple items here? Are they just going to be comma separated? Overall, I think it's ok to just mirror what Resend and others do because we don't want to reinvent the interface, but just curious.

artahian (approved):
Approving with optional comments

packages/aws-ses/README.md
omegascorp: I've added it since we will likely use React to store templates for our emails, I think it will be a convenient API. It's natively supported only by Resend.

PR #43, opened 2025-07-18
Improved config

omegascorp:
* Support of booleans in local config
* Support of numbers in local config
* New config variables for email support

### DIFF
--- a/packages/modelence/src/auth/user.ts
+++ b/packages/modelence/src/auth/user.ts
@@ -52,6 +52,21 @@ export default new Module('_system.user', {
     limit: 200,
   }],
   configSchema: {
+    'auth.email.enabled': {
+      type: 'boolean',
+      isPublic: true,
+      default: true,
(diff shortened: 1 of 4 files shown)


packages/modelence/src/auth/user.ts
```diff
@@ -52,6 +52,21 @@ export default new Module('_system.user', {
...
   }],
   configSchema: {
+    'auth.email.enabled': {
+      type: 'boolean',
+      isPublic: true,
+      default: true,
+    },
+    'auth.email.from': {
+      type: 'string',
+      isPublic: true,
```
artahian: Does this need to be public?

artahian (approved):

packages/modelence/src/auth/user.ts
omegascorp: I will add those later, the scheme might change

PR #44, opened 2025-07-18
Require also application alias when deploying application to cloud

rajiv-pat:
We should require also application alias when deploying application to cloud. New deployment command will be "npx modelence deploy --app appAlias --env envAlias".
This way we will define application in a unique way and if user will approve the deployment it will be deployed for the correct {application, environment} pair.

### DIFF
--- a/packages/modelence/src/bin/deploy.ts
+++ b/packages/modelence/src/bin/deploy.ts
@@ -5,7 +5,7 @@ import { authenticateCli } from './auth';
 import { getStudioUrl, getBuildPath, getProjectPath } from './config';
 import { build } from './build';
 
-export async function deploy(options: { env: string }) {
+export async function deploy(options: { app: string; env: string }) {
   const cwd = process.cwd();
   const modelenceDir = join(cwd, '.modelence');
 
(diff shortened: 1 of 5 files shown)


packages/modelence/src/bin/deploy.ts
```diff
@@ -97,12 +97,17 @@ async function createBundle(bundlePath: string) {
   console.log(`Deployment bundle created at: ${bundlePath} (${(stats.size / 1024 / 1024).toFixed(2)} MB)`);
 }
 
-async function uploadBundle(environmentAlias: string, bundlePath: string, token: string) {
-  const response = await fetch(getStudioUrl(`/api/environments/${environmentAlias}/upload`), {
+async function uploadBundle(appAlias: string, envAlias: string, bundlePath: string, token: string) {
+  const response = await fetch(getStudioUrl(`/api/upload`), {
```
artahian: Can we rename this to `upload-bundle`? Now that we've removed the rest of the path `/api/upload` is too generic and could refer to any upload (in the future we can have other assets to upload).

artahian (commented):
All good, just one comment

packages/modelence/src/bin/deploy.ts
rajiv-pat: sure, done ✅

artahian (approved):

PR #45, opened 2025-07-19
MOD-53: create-modelence-app - fetch examples zip instead of using GitHub API

omegascorp:
* Use  fetch zip of `examples` repository and then unzip it instead of using api


rajiv-pat (commented):
@omegascorp but user should have git installed on their environment, right? So we force them to have "git" available.

artahian:
@omegascorp can we check and replace how the most common create app tools do this, like `create-next-app`?

rajiv-pat:
@omegascorp, it seems there is a merge issue and an npm publish issue. Could you please fix and merge this branch? also publish it..

PR #46, opened 2025-07-23
MOD-29: Email confirmation

omegascorp:
* Send email confirmation on signup.
* API endpoint to confirm email.
* After email confirmation, redirects to `emailVerifiedRedirectUrl` which is passed to startApp

Note: the email confirmation is optional at the moment, we can add option to make it mandatory, but I suggest doing it in a separate PR


packages/auth-ui/src/components/LoginForm.tsx
```diff
@@ -32,6 +32,7 @@ export function LoginForm({
   inputClassName = "",
   labelClassName = ""
 }: LoginFormProps) {
+  const isEmailAuthEnabled = getConfig('_system.user.auth.email.enabled');
```
artahian: Is there ever a case when this might not be enabled? i.e. what's a use case where one would force using social logins and no email/password option?

packages/modelence/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.5.7-dev.6",
+  "version": "0.5.7-dev.7",
```
artahian: FYI I believe @rajiv-pat published dev.7 already earlier

packages/modelence/src/auth/client/index.ts
```diff
@@ -41,6 +41,22 @@ export async function loginWithPassword(options: { email: string, password: stri
...
+ * @example
+ * ```ts
+ * await verifyEmail({ token: 'verification-token' });
+ * ```
+ * @param options.token - The email verification token.
+ */
+export async function verifyEmail(options: { token: string }) {
+  const { token } = options;
+  const { user } = await callMethod<{ user: UserInfo }>('_system.user.verifyEmail', { token });
+  setCurrentUser(user);
```
artahian: I'm not sure about having this method auto-login the user - is this a standard behavior in general?

packages/modelence/src/auth/db.ts
```diff
@@ -53,3 +53,24 @@ export const dbDisposableEmailDomains = new Store('_modelenceDisposableEmailDoma
     }
   ]
 });
+
+export const tokensCollection = new Store('_modelenceTokens', {
```
artahian: This collection name is too generic, there may be other tokens for different purposes. Can we rename it to `_modelenceEmailVerificationTokens`? We can use a shorter name too but so that it is still clear that this is for email verification.

packages/modelence/src/auth/signup.ts
```diff
@@ -58,7 +61,36 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
...
   });
 
-  // TODO: send verification email
+  if (getEmailConfig().provider) {
+    const emailProvider = getEmailConfig().provider;
+    const baseUrl = connectionInfo?.baseUrl;
+    
+    // Generate verification token
+    const verificationToken = randomBytes(32).toString('hex');
+    const expiresAt = new Date(Date.now() + 24 * 60 * 60 * 1000); // 24 hours from now
```
artahian: Can we use the time utility here? We already have time.ts in modelence that provides convenient functions for things like this.

packages/modelence/src/auth/db.ts
artahian: I see that it has `type` to distinguish, but mixing different purposes into one collection is usually not a good idea because not all fields will apply to all and it becomes hard to maintain (i.e. not all tokens might be related to a userId / email).

packages/modelence/src/auth/signup.ts
```diff
@@ -58,7 +61,36 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
...
+    await tokensCollection.insertOne({
+      userId: result.insertedId,
+      type: 'emailVerification',
+      email,
+      token: verificationToken,
+      createdAt: new Date(),
+      expiresAt,
+    });
+    
+    const verifyUrl = `${baseUrl}/email-verification?token=${verificationToken}`;
```
artahian: I think we need some sort of a prefix here because we might run into collisions with user-defined routes as we keep adding our built-in routes.

For example for api routes like method calls, we use `/api/_internal/...`. Since this is not an API, we shouldn't use the `/api` prefix, and also `_internal` doesn't look good on routes that users can see. But maybe we can at least add special prefix groups like `/_auth/email-verification`? Later we can use the `_auth` prefix for other built-in auth related routes like reset password / forgot password, and introduce other prefix groups for other things that start with underscore also to avoid collisions.

packages/modelence/src/auth/verifyEmail.ts
```diff
@@ -0,0 +1,68 @@
+import { z } from 'zod';
+
+import { Args, Context } from '../methods/types';
+import { usersCollection, tokensCollection } from './db';
+import { setSessionUser } from './session';
+
+export async function handleVerifyEmail(args: Args, { session }: Context) {
+  if (!session) {
+    throw new Error('Session is not initialized');
+  }
```
artahian: This can never happen, right?

packages/modelence/src/auth/verifyEmail.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  if (userDoc.emails[emailIndex].verified) {
+    throw new Error('Email is already verified');
+  }
+
+  // Mark the specific email as verified
+  await usersCollection.updateOne(
+    { _id: tokenDoc.userId },
+    { $set: { [`emails.${emailIndex}.verified`]: true } }
+  );
```
artahian: This part looks very flaky, and it is also very vulnerable to a race condition (if the emails list changes in between these operations everything will get mixed up).

MongoDB should already have a way to find & update a matching element in an array in a single command.
https://www.mongodb.com/community/forums/t/update-nested-sub-document-with-a-specific-condition/136373

packages/auth-ui/src/components/LoginForm.tsx
omegascorp: It might be useful, but it's a rare case, I guess. If we move the settings this can be removed too

packages/modelence/package.json
omegascorp: That's fine, will bump one more time

packages/modelence/src/auth/client/index.ts
omegascorp: No, I suppose to remove it and forgot. There is a case to use it as a vulnerability - you can sign in without knowing the password if you overtake the email.

packages/modelence/src/auth/db.ts
omegascorp: Ok, I will split it

packages/modelence/src/auth/signup.ts
omegascorp: This is a user-defined route, it should be in the app itself since they need to be able to control the UI. And they will be able to change the URL too once they will be able to control email template

packages/modelence/src/auth/verifyEmail.ts
omegascorp: Currently, the session is defined as `session: Session | null;` if it can't happen, we need to change the type.

artahian:
@omegascorp is this ready for another review? I noticed you already made these changes but haven't re-requested the review.

artahian (approved):

PR #48, opened 2025-07-27
MOD-51: AI telemetry missing fields

artahian:


packages/ai/src/index.ts
```diff
@@ -89,7 +89,21 @@ export async function generateText(options: GenerateTextOptions) {
...
+          usage: {
+            promptTokens: result.usage.promptTokens,
+            completionTokens: result.usage.completionTokens,
+            totalTokens: result.usage.totalTokens,
+          }
+        }
+      });
+    } else {
+      // Backwards compatibility for older versions of Modelence
+      // @ts-ignore
```
omegascorp, line 104: I'm to sure why there is @ts-ignore?

artahian: This is because TypeScript thinks this "else" should never happen, and it derives these types as "never" based on the current version of Modelence, giving errors. But the "else" happens when the Modelence version is older, which TypeScript doesn't know about.

PR #49, opened 2025-07-27
fix google auth

omegascorp:
Turned out I borke Google Authentication while I was updating the format of config.


rajiv-pat (approved):

rajiv-pat:
approving, but seems that there might be other issues with Google Auth, so need to double check.

PR #51, opened 2025-07-29
Update types in modelece to 1.0.3

omegascorp:
Update @modelence/types in modelence to 1.0.3


omegascorp:
@rajiv-pat I'm merging this without approval, it's a chore chane anyway

PR #55, opened 2025-08-01
MOD-66: Email verification template

omegascorp:
* Allow passing email verification template to startApp

### DIFF
--- a/packages/modelence/src/app/emailConfig.ts
+++ b/packages/modelence/src/app/emailConfig.ts
@@ -12,7 +12,7 @@ export type EmailConfig = {
     }) => string;
     redirectUrl?: string;
   };
-  // deprecated
+  // @deprecated use verification.redirectUrl instead
   emailVerifiedRedirectUrl?: string;
 };
 
--- a/packages/modelence/src/app/emailConfig.ts
+++ b/packages/modelence/src/app/emailConfig.ts
@@ -1,8 +1,18 @@
 import { EmailProvider } from "@modelence/types";
 
-type EmailConfig = {
+export type EmailConfig = {
   provider?: EmailProvider;
   from?: string;
+  verification?: {
+    subject?: string;
+    template?: (props: {
+      name: string;
+      email: string;
+      verificationUrl: string;
+    }) => string;
+    redirectUrl?: string;
+  };
+  // deprecated
   emailVerifiedRedirectUrl?: string;
 };
 
--- a/packages/modelence/src/auth/signup.ts
+++ b/packages/modelence/src/auth/signup.ts
@@ -73,8 +73,8 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
 
   if (getEmailConfig().provider) {
     const emailProvider = getEmailConfig().provider;
-    const baseUrl = connectionInfo?.baseUrl;
-    
+    const baseUrl = process.env.MODELENCE_SITE_URL || connectionInfo?.baseUrl;
+
     // Generate verification token
     const verificationToken = randomBytes(32).toString('hex');
     const expiresAt = new Date(Date.now() + time.hours(24));
--- a/packages/modelence/src/auth/signup.ts
+++ b/packages/modelence/src/auth/signup.ts
@@ -12,9 +12,9 @@ import { time } from '@/time';
 
 function defaultEmailVerificationTemplate({ name, email, verificationUrl }: { name?: string; email: string; verificationUrl: string }) {
   return `
-    <p>Hello${name ? ` ${name}` : ''},</p>
+    <p>Hi${name ? ` ${name}` : ''},</p>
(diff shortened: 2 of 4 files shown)


packages/modelence/src/app/emailConfig.ts
```diff
@@ -1,8 +1,18 @@
 import { EmailProvider } from "@modelence/types";
 
-type EmailConfig = {
+export type EmailConfig = {
   provider?: EmailProvider;
```
artahian, line 4: @omegascorp probably not a blocker right now, but I don't think we have any docs around this - can we add documentation on how email config works? Typedoc comments + adding an "Email" section to our Mintlify docs should do it.

packages/modelence/src/auth/signup.ts
```diff
@@ -10,6 +10,15 @@ import { getConfig } from '@/server';
 import { getEmailConfig } from '@/app/emailConfig';
 import { time } from '@/time';
 
+function defaultEmailVerificationTemplate({ name, email, verificationUrl }: { name?: string; email: string; verificationUrl: string }) {
+  return `
+    <p>Hello${name ? ` ${name}` : ''},</p>
```
artahian:
```suggestion
    <p>Hi${name ? ` ${name}` : ''},</p>
```

For some reason "Hello [name]" sounds almost like spam to me, I believe most platforms just use "Hi [name]"

packages/modelence/src/auth/signup.ts
```diff
@@ -10,6 +10,15 @@ import { getConfig } from '@/server';
...
+    <p>Please verify your email address ${email} by clicking the link below:</p>
+    <p><a href="${verificationUrl}">Verify Email</a></p>
```
artahian: Can we also use the `verificationUrl` as the text of the link? I think it's standard to display the URL itself so the user can easily see where it is leading and also copy/paste it if the link somehow is not rendered / doesn't work.

packages/modelence/src/auth/signup.ts
```diff
@@ -64,8 +73,8 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
 
   if (getEmailConfig().provider) {
     const emailProvider = getEmailConfig().provider;
-    const baseUrl = connectionInfo?.baseUrl;
-    
+    const baseUrl = process.env.MODELENCE_SITE_URL || connectionInfo?.baseUrl;
```
artahian, line 76: Where is `MODELENCE_SITE_URL` coming from? I believe @rajiv-pat introduced `ROOT_URL` somewhere else for this.

packages/modelence/src/auth/signup.ts
```diff
@@ -81,17 +90,20 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
     
     const verifyUrl = `${baseUrl}/api/_internal/auth/verify-email?token=${verificationToken}`;
     
+    const template = getEmailConfig()?.verification?.template || defaultEmailVerificationTemplate;
+    // TODO: we should have also the name on this step
+    const htmlTemplate = template({ name: '', email, verificationUrl: verifyUrl });
+    const textContent = htmlTemplate.replace(/<[^>]*>/g, '').replace(/\s+/g, ' ').trim();
```
artahian, line 96: I'm not sure what exactly this does, but is there a way to do it without a custom regex?

packages/modelence/src/auth/signup.ts
omegascorp: ROOT_URL is used only in the studio. I'm planning to allow adding it to the modelce config (let me know what do you think). I didn't use ROOT_URL for 2 reasons:
It does not have MODELENCE prefix
It's not common, for nextjs it's NEXT_PUBLIC_SITE_URL, based on chat gpt SITE_URL is the most common variable I can dig dipper if you need more alternatives

Now when I review it, I think this should be probably also added to config, so the users can edit it from the cloud

packages/modelence/src/auth/signup.ts
omegascorp: It removes text in <> (HTML) and replaces s+ (double+ spaces) with a single space. I can search for a lib that does this, but I didn't want to add an additional dependency

packages/modelence/src/auth/signup.ts
omegascorp: @artahian is it ok if we define `_system.siteUrl` as a global config? The thing is it seems like we didn't define config shemas for the global configs like `_system.elastic.apiKey`, so when it's not set, there is no way to add it in the UI

artahian: @omegascorp I'm ok with adding it as a global config, but _system configs are not editable on purpose (we don't want someone to accidentally mess up their elastic URL for example and not be able to recover).

PR #57, opened 2025-08-06
modelence, auth-ui: Make email verification mandatory

omegascorp:
* Send email verification on sign-up
* Send email verification on sign-in if the email is not verified, and don't allow them to sign in
* If there is no email provider, it allows signing in without confirmation. I think this will work well for the initial setup when the email is not set up yet

Screenshot 2025-08-06 at 11 07 52

Screenshot 2025-08-06 at 11 16 33

### DIFF
--- a/packages/auth-ui/src/components/SignupForm.tsx
+++ b/packages/auth-ui/src/components/SignupForm.tsx
@@ -1,5 +1,5 @@
 import { getConfig, signupWithPassword } from 'modelence/client';
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { GoogleIcon } from './icons/GoogleIcon';
 import { Button } from './ui/Button';
 import { Card, CardContent, CardFooter, CardHeader, CardTitle } from './ui/Card';
@@ -83,6 +83,7 @@ export function SignupForm({
   labelClassName = "",
   consents
 }: SignupFormProps) {
+  const [isSignupSuccess, setIsSignupSuccess] = useState(false);
   const isGoogleAuthEnabled = getConfig('_system.user.auth.google.enabled');
 
   const handleSubmit = useCallback(async (event: React.FormEvent<HTMLFormElement>) => {
@@ -98,8 +99,13 @@ export function SignupForm({
       return;
     }
     
-    await signupWithPassword({ email, password });
-  }, []);
+    try {
+      await signupWithPassword({ email, password });
+      setIsSignupSuccess(true);
+    } catch (error) {
+      onError(error as Error);
+    }
+  }, [onError]);
 
   const openGoogleAuth = useCallback(() => {
     window.location.href = '/api/_internal/auth/google';
@@ -123,6 +129,27 @@ export function SignupForm({
     }
     return buttons;
   }, []);
+
+  if (isSignupSuccess) {
+    return (
+      <Card className={`w-full max-w-md mx-auto bg-white dark:bg-gray-900 text-gray-900 dark:text-white ${cardClassName}`}>
+        <CardHeader className="text-center">
+          <CardTitle className="text-xl">
+            Check your email
+          </CardTitle>
+        </CardHeader>
+        
+        <CardContent className="text-center space-y-4">
+          <p className="text-gray-600 dark:text-gray-400">
+            We've sent you a verification email. Please check your inbox and click the verification link to activate your account.
+          </p>
+          <p className="text-sm text-gray-500 dark:text-gray-500">
+            Don't see the email? Check your spam folder.
+          </p>
+        </CardContent>
+      </Card>
+    );
+  }
 
   return (
     <Card className={`w-full max-w-md mx-auto bg-white dark:bg-gray-900 text-gray-900 dark:text-white ${cardClassName}`}>
--- a/packages/modelence/src/auth/login.ts
+++ b/packages/modelence/src/auth/login.ts
@@ -5,6 +5,7 @@ import { Args, Context } from '../methods/types';
 import { usersCollection } from './db';
 import { clearSessionUser, setSessionUser } from './session';
 import { sendVerificationEmail } from './verification';
+import { getEmailConfig } from '@/app/emailConfig';
 
 export async function handleLoginWithPassword(args: Args, { user, session, connectionInfo }: Context) {
   if (!session) {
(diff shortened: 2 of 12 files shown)


packages/auth-ui/src/components/SignupForm.tsx
```diff
@@ -124,6 +130,27 @@ export function SignupForm({
...
+            Check your email
+          </CardTitle>
+        </CardHeader>
+        
+        <CardContent className="text-center space-y-4">
+          <p className="text-gray-600 dark:text-gray-400">
+            We've sent you a verification email. Please check your inbox and click the verification link to activate your account.
+          </p>
+          <p className="text-sm text-gray-500 dark:text-gray-500">
+            Don't see the email? Check your spam folder.
```
artahian, line 147: Can we add a "Resend" option also? I think it's pretty standard and otherwise if the user somehow loses the email or it expires they will be stuck.

artahian: Also, should we just give users control on how to handle the verification state? I'm worried that there is too much magic happening in the SignupForm component now and it covers the email verification state as well. Can we check how Supabase does this for example (https://supabase.com/docs/guides/auth/auth-helpers/auth-ui), or https://www.better-auth.com?

packages/modelence/src/auth/login.ts
```diff
@@ -19,13 +21,27 @@ export async function handleLoginWithPassword(args: Args, { user, session }: Con
...
     { 'emails.address': email },
     { collation: { locale: 'en', strength: 2 } }
   );
 
-  const passwordHash = userDoc?.authMethods?.password?.hash;
+  if (!userDoc) {
+    throw incorrectCredentialsError();
+  }
+
+  const emailDoc = userDoc?.emails?.find(e => e.address === email);
```
artahian:
```suggestion
  const emailDoc = userDoc.emails?.find(e => e.address === email);
```

packages/auth-ui/src/components/SignupForm.tsx
omegascorp: @artahian I think it should be fine for the auth-ui. Supabase deprecated their auth-ui, but it also does some magic, you just put the component and it works

artahian: @omegascorp I guess that part is ok, but what about the resend part?

Right now, if you don't click on the email before it expires or somehow don't receive the email (or accidentally delete it), you seem to be forever locked as a user, and can't log in, but at the same time you also can't sign up with that same email anymore.

I think at the minimum we need to allow users to request the verification email again. Perhaps we should add a separate verification page (or this can be a different state of the login form component) that shows the same card, saying something like:

"You need to verify your email before logging in. We've sent an email to xxx@yyy.zz, click the link, etc, etc. 

[Resend Button]"

this component can be used for both cases, at any time. After signup, we can redirect to the verification page (or to the login form, but immediately to this state that shows the verification card). And at any time, if the user later comes back in and tries to log in, they will see the same card, with the same Resend button option.

This way we have a single consistent card, regardless of whether it's right after signup or coming back later, and there will always be a way to resend the email without being completely locked out.

omegascorp: > Send email verification on sign-in if the email is not verified, and don't allow them to sign in

Aram Jan, please check the second screenshot. It sends an email verification automatically when you are trying to sign in with an email that is not verified. Please let me know if this aproach is ok, this way was requiring the minimal changes and will not require users to make their UI for this case if they are not using auth-ui

artahian: I think everything in the screenshot is totally ok, and we can do both at the same time to be a win/win - let's discuss this live to clarify (just pinged you).

omegascorp:
@artahian, I've added a rate limit on login and a separate one on verification. I've also moved sign-up verification email up, since without that you can, for example, brute-force which emails we consider disposable, or check which emails are registered.

packages/modelence/src/auth/signup.ts
```diff
@@ -13,6 +13,15 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
...
 
+  const ip = connectionInfo?.ip;
+  if (ip) {
+    await consumeRateLimit({
+      bucket: 'signup',
+      type: 'ip',
+      value: ip,
+    });
+  }
+
```
artahian: @omegascorp but I think this is going to be an issue if you keep signing up with existing emails and run out of attempts without even having signed up a single user. The original rate limit's purpose was different - to limit the number of new users created by the same person.

Can we keep both? I think this new rate limit is also useful, but we can call it something like `signupAttempt` instead, and have a different/higher limit on it, so we can separately control both attempts and actual new users.

packages/modelence/src/auth/user.ts
```diff
@@ -59,6 +59,26 @@ export default new Module('_system.user', {
...
+    bucket: 'signin',
+    type: 'ip',
+    window: time.minutes(15),
+    limit: 20,
+  }, {
+    bucket: 'signin',
+    type: 'ip',
+    window: time.days(1),
+    limit: 200,
+  }, {
```
artahian: Shouldn't signin limits be higher than signup? Maybe 50/500?

packages/modelence/src/auth/user.ts
```diff
@@ -59,6 +59,26 @@ export default new Module('_system.user', {
...
+    bucket: 'verification',
```
artahian, line 83: Shouldn't the bucket be the user for the verification, since we have the user at that point already?

packages/modelence/src/auth/login.ts
```diff
@@ -19,13 +31,39 @@ export async function handleLoginWithPassword(args: Args, { user, session }: Con
...
+
+    await sendVerificationEmail({
+      userId: userDoc?._id,
+      email,
+      baseUrl: connectionInfo?.baseUrl,
+    });
+    throw new Error("Your email address hasn't been verified yet. We've sent a new verification email to your inbox.");
+  }
+
+  const passwordHash = userDoc.authMethods?.password?.hash;
```
artahian: Shouldn't this check be before the email verified checks? If the user is providing a wrong password, we should just reject the login, right? Otherwise you can still brute force unverified emails and trigger the verification emails even though you don't even have the password.

packages/modelence/src/auth/signup.ts
```diff
@@ -71,38 +59,19 @@ export async function handleSignupWithPassword(args: Args, { user, connectionInf
...
-      from: getEmailConfig()?.from || 'noreply@modelence.com',
-      subject: getEmailConfig()?.verification?.subject || 'Verify your email address',
-      text: textContent,
-      html: htmlTemplate,
+  if (ip) {
+    await consumeRateLimit({
+      bucket: 'signup',
+      type: 'ip',
+      value: ip,
     });
```
artahian: I think this should still be in its original position, before inserting the user document? Otherwise we still do the signup, create the user, and only then check the rate limit, which is too late.

omegascorp:
@artahian comments are addressed

PR #60, opened 2025-08-07
Add client/public folder to vite server 

omegascorp:


artahian (commented):
As discussed live, it sounds like we both agree on the public folder idea, let's just read it from the project's own vite config file.

packages/modelence/src/viteServer.ts
```diff
@@ -26,7 +28,8 @@ class ViteServer implements AppServer {
       return (this.viteServer?.middlewares ?? []) as ExpressMiddleware[];
     }
     
-    return [express.static('./.modelence/build/client'), express.static('./src/client/public')];
+    const publicDir = this.config?.publicDir || './src/client/public';
```
artahian: When the public dir is not explicitly specified, can we just not publish it at all instead of having a default path? Because the default makes too many assumptions (i.e. the fact that there's an `src` folder, etc, which all depends on other config params).

PR #61, opened 2025-08-09
next: Make next a peer dependency

omegascorp:
* Make next a peer dependency so it's version can be controlled from the app


artahian (approved):
Shouldn't it also be a dev dependency at the same time?

PR #62, opened 2025-08-16
modelence, auth-ui: Password Reset

omegascorp:
* Add Password Reset UI
* Add Password Reset Methods

Note: the usages should also be updated in the example apps, since the reset password requires new routes. It's backwards compatable so it should not break existing apps.

### DIFF
--- a/packages/auth-ui/package-lock.json
+++ b/packages/auth-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelence/auth-ui",
-  "version": "0.1.16-dev.0",
+  "version": "0.1.16",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelence/auth-ui",
-      "version": "0.1.16-dev.0",
+      "version": "0.1.16",
       "license": "ISC",
       "dependencies": {
         "@modelence/react-query": "^1.0.0",
--- a/packages/auth-ui/package.json
+++ b/packages/auth-ui/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "@modelence/auth-ui",
-  "version": "0.1.16-dev.0",
+  "version": "0.1.16",
   "description": "Authentication UI components for Modelence",
   "exports": {
     ".": {
--- a/packages/aws-ses/package-lock.json
+++ b/packages/aws-ses/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelence/aws-ses",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelence/aws-ses",
-      "version": "1.0.0",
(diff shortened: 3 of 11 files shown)


packages/modelence/src/auth/resetPassword.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // Find user by email
+  const userDoc = await usersCollection.findOne(
+    { 'emails.address': email },
+    { collation: { locale: 'en', strength: 2 } }
+  );
+
+  if (!userDoc) {
+    // For security, don't reveal if email exists or not
+    return { success: true, message: 'If an account with that email exists, a password reset link has been sent' };
```
rajiv-pat: Seems that we are using the same phrase in 3 different places. Let's keep it as a const somewhere.
`If an account with that email exists, a password reset link has been sent`

packages/modelence/src/auth/db.ts
```diff
@@ -73,3 +73,22 @@ export const emailVerificationTokensCollection = new Store('_modelenceEmailVerif
     }
   ]
 });
+
+export const resetPasswordTokensCollection = new Store('_modelenceResetPasswordTokens', {
```
rajiv-pat, line 77: Do we really need a new collection for password reset tokens? Can't we keep them in the users document?

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -0,0 +1,133 @@
...
+    html: htmlTemplate,
+  });
+
+  return { success: true, message: 'If an account with that email exists, a password reset link has been sent' };
+}
+
+export async function handleResetPassword(args: Args, { }: Context) {
+  const token = z.string().parse(args.token);
+  const password = z.string()
+    .min(8, { message: 'Password must contain at least 8 characters' })
```
rajiv-pat: This part seems to be a copy a paste from the handleSignupWithPassword. Let's move it to some common place, otherwise if we will decide to do some change in the password rules, we can forget to change it for the password reset.

rajiv-pat (changes requested):
Left some comments

packages/modelence/src/auth/db.ts
omegascorp: With separate collection, we can use expireAfterSeconds to delete the document when it expires automatically. Since it should not always be fetched with the user, I don't see a reason to make it a subdocument (we don't get any benefits from it).

omegascorp:
@eduard-piliposyan the feedback is applied please check again

packages/modelence/src/auth/login.ts
```diff
@@ -22,7 +23,8 @@ export async function handleLoginWithPassword(args: Args, { user, session, conne
     });
   }
 
-  const email = z.string().email().parse(args.email);
+  const email = validateEmail(args.email as string);
+  // password is accepted just as a string, so users can still sign in if the password validation rules are changes
```
rajiv-pat: grammatical typo. Should be `... validation rules are changed`

Issue #26, opened 2025-06-23
Make MongoDB connection more resilient


artahian:
Currently, things like upgrading the MongoDB Atlas cluster from Serverless to Dedicated seems to permanently break the database connection and requires restarting the app, which is too bad.

We should make the MongoDB connection self healing, so regardless of what happens to the cluster (failover, tier upgrade, temporary outage, etc), as long as it's back the app automatically reconnects and continues working like nothing happened.

This should be as simply as periodically testing the db connection and doing exponential backoff reconnect attempts if the test fails.

omegascorp:
@artahian, note: this seems to be an edge case that is not actually reconnecting. Upgrading Serverless -> Dedicated likely means that you have changed the physical DB. MongoClient is actually reconnected, but didn't manage to find the same user since likely in the new DB it's a new user with the same credentials (but different id)

omegascorp:
Closing the issue as we decided to not handle this case for now

PR #63, opened 2025-08-20
fix(db): ensure MongoDB Retries to connect db making its connection self healing

AtulYadav25:
Fixed Issue #26 

## With this Update
- Updated `client.ts` to retry connection with mongodb when connection lost.
- Max Retry of 10 Attempts to connect to mongodb
- Verified locally with a test script.

### Testing
- Ran `npx tsx test-db.js` successfully after the change.

test-db.js code below
```
import { connect } from "./src/db/client";

async function run() {
  const client = await connect();
  const db = client.db("testdb");

  setInterval(async () => {
    try {
      await db.command({ ping: 1 });
      console.log("✅ Ping succeeded");
    } catch (err) {
      // console.error("❌ Ping failed", err);
    }
  }, 2000);
}

run();

```

Used Docker to run mongodb locally and connect to `packages/modelence/src/db/client.ts` with custom mongodb URI of docker mongodb

**Test Case 1: Mongodb Is running on docker**
Screenshot (159)
Succeded✔

**Test Case 2: Mongodb Stopped on docker**
Screenshot (160)
Succeded✔

**Test Case 3: Mongodb Restart On Docker**
Screenshot (161)
Succeded✔

### DIFF
--- a/packages/modelence/src/db/client.ts
+++ b/packages/modelence/src/db/client.ts
@@ -52,10 +52,7 @@ export async function closeConnection() {
     healthCheckInterval = null;
   }
   
-  if (client) {
-    await client.close();
-    client = null;
-  }
+  await closeExistingClientConnection();
   
   reconnecting = false;
 }
--- a/packages/modelence/src/db/client.ts
+++ b/packages/modelence/src/db/client.ts
@@ -2,8 +2,10 @@ import { MongoClient, ServerApiVersion } from 'mongodb';
 import { getConfig } from '../config/server';
 
 let client: MongoClient | null = null;
+let reconnecting = false;
+let healthCheckInterval: number | null = null;
 
-export async function connect() {
+export async function connect(): Promise<MongoClient> {
   if (client) return client;
 
   const mongodbUri = getMongodbUri();
@@ -12,7 +14,10 @@ export async function connect() {
   }
 
   client = new MongoClient(mongodbUri, {
-    maxPoolSize: 20
+    maxPoolSize: 20,
+    retryWrites: true,
+    serverApi: ServerApiVersion.v1,
+    serverSelectionTimeoutMS: 5000,
   });
 
   try {
@@ -21,6 +26,8 @@ export async function connect() {
     // Send a ping to confirm a successful connection
     await client.db("admin").command({ ping: 1 });
     console.log("Pinged your deployment. You successfully connected to MongoDB!");
+
+    startHealthCheck(); // begin self-healing loop
     return client;
   } catch (err) {
     console.error(err);
@@ -38,9 +45,77 @@ export function getClient() {
   return client;
 }
 
-// export async function closeConnection() {
-//   if (client) {
-//     await client.close();
-//     client = null;
-//   }
-// }
+export async function closeConnection() {
+  if (healthCheckInterval) {
+    clearInterval(healthCheckInterval);
+    healthCheckInterval = null;
+  }
+  
+  if (client) {
+    await client.close();
+    client = null;
+  }
+  
+  reconnecting = false;
+}
+
+async function attemptReconnect() {
+  if (reconnecting) return;
+  reconnecting = true;
+
+  let retries = 0;
+  const maxRetries = 10;
+  
+  while (retries < maxRetries) {
+    try {
+      retries++;
+      console.log(`Attempting MongoDB reconnect (#${retries}/${maxRetries})...`);
+      
+      // Close existing client if it exists
+      if (client) {
+        try {
+          await client.close();
+        } catch (closeErr) {
+          console.warn('Error closing existing client:', closeErr);
+        }
+        client = null;
+      }
+      
+      await connect();
+      console.log("Reconnected to MongoDB");
+      reconnecting = false;
+      return;
+    } catch (err) {
+      const delay = Math.min(30000, 1000 * Math.pow(2, retries)); //Max delay of 30s
+      console.error(
+        `Reconnect attempt #${retries}/${maxRetries} failed: ${err}. Retrying in ${delay / 1000}s...`
+      );
+      
+      if (retries < maxRetries) {
+        await new Promise((res) => setTimeout(res, delay));
+      }
+    }
+  }
+  
+  console.error(`Failed to reconnect to MongoDB after ${maxRetries} attempts`);
+  reconnecting = false;
+}
+
+function startHealthCheck() {
+  // Clear existing health check to prevent duplicates
+  if (healthCheckInterval) {
+    clearInterval(healthCheckInterval);
+  }
+  
+  healthCheckInterval = setInterval(async () => {
+    if (!client) return;
+    
+    try {
+      await client.db("admin").command({ ping: 1 });
+    } catch (err) {
+      console.error("MongoDB ping failed, attempting to reconnect...", err);
+      client = null;
+      attemptReconnect();
+    }
+  }, 15000); // check every 15s
+}
--- a/packages/modelence/src/db/client.ts
+++ b/packages/modelence/src/db/client.ts
@@ -1,40 +1,51 @@
 import { MongoClient, ServerApiVersion } from 'mongodb';
 import { getConfig } from '../config/server';
 import { time } from '../time';
+import { EventEmitter } from "events";
 
 let client: MongoClient | null = null;
 let reconnecting = false;
-let healthCheckInterval: number | null = null;
+let healthCheckInterval: NodeJS.Timeout | null = null;
+let connectingPromise: Promise<MongoClient> | null = null;
+export const dbEvents = new EventEmitter();
 
 export async function connect(): Promise<MongoClient> {
   if (client) return client;
+
+  if (connectingPromise) return connectingPromise;
 
   const mongodbUri = getMongodbUri();
   if (!mongodbUri) {
     throw new Error('MongoDB URI is not set');
   }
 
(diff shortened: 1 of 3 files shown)


packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+        }
+        client = null;
+      }
+      
+      await connect();
+      console.log("Reconnected to MongoDB");
+      reconnecting = false;
+      return;
+    } catch (err) {
+      const delay = Math.min(30000, 1000 * Math.pow(2, retries)); //Max delay of 30s
```
omegascorp: please use time.seconds(30) and time.seconds(1) from 'modelence/server'

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+  }
+  
+  healthCheckInterval = setInterval(async () => {
+    if (!client) return;
+    
+    try {
+      await client.db("admin").command({ ping: 1 });
+    } catch (err) {
+      console.error("MongoDB ping failed, attempting to reconnect...", err);
+      client = null;
```
omegascorp: It seems like resettingthe  client is extra here, since its done also in the `attemptReconnect`

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+
+  let retries = 0;
+  const maxRetries = 10;
+  
+  while (retries < maxRetries) {
+    try {
+      retries++;
+      console.log(`Attempting MongoDB reconnect (#${retries}/${maxRetries})...`);
+      
+      // Close existing client if it exists
```
omegascorp, line 98: Can you move the closing of the client to a separate function?

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+  
+  healthCheckInterval = setInterval(async () => {
+    if (!client) return;
+    
+    try {
+      await client.db("admin").command({ ping: 1 });
+    } catch (err) {
+      console.error("MongoDB ping failed, attempting to reconnect...", err);
+      client = null;
+      attemptReconnect();
```
omegascorp: Could you call it just `reconnect`?

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+    }
+  }, 15000); // check every 15s
```
omegascorp: Please replace with `time.seconds(15)`

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +45,77 @@ export function getClient() {
...
+  if (reconnecting) return;
+  reconnecting = true;
+
+  let retries = 0;
+  const maxRetries = 10;
+  
+  while (retries < maxRetries) {
+    try {
+      retries++;
+      console.log(`Attempting MongoDB reconnect (#${retries}/${maxRetries})...`);
```
omegascorp: Could you replace `Attempting MongoDB reconnect` with `Reconnecting to MongoDB`?

omegascorp (commented):
@AtulYadav25 thank you for making this PR. I've added some comments to it. Please let me know if you can address them. Also, please don't hesitate to let me know if you disagree with any of the suggestions.

packages/modelence/src/db/client.ts
AtulYadav25: Thanks, I’ve updated this to remove the  client = null in startHealthCheck + extracted closeExistingClientConnection() + improved logs as suggested.

packages/modelence/src/db/client.ts
```diff
@@ -38,9 +46,81 @@ export function getClient() {
...
+export async function closeConnection() {
+  if (healthCheckInterval) {
+    clearInterval(healthCheckInterval);
+    healthCheckInterval = null;
+  }
+  
+  if (client) {
+    await client.close();
+    client = null;
+  }
```
omegascorp: One more thing, could you use `closeExistingClientConnection` here?

AtulYadav25: Updated ✅! It's Great how you simplified it

AtulYadav25: Are we going ahead with this PR?

omegascorp: @AtulYadav25 Sorry for the delay, there is a technical issue in the PR, since the is a getClient method, the reference to the current client might be saved in different places, and after creating a new instance. It's actually what currently happens with stores. All new queries from stores basically fail after reconnecting. I was delaying the response to suggest a proper solution that I don't have yet. Please let me know if you have any suggestions on how to resolve this issue.

AtulYadav25: From your comment, what I understood is that in `startApp()` func, the initStores function is only called once. For example, when the app first connects to DB1, the stores are initialized with that client reference. But when MongoDB reconnects and a new client is created, initStores is not called again — so the stores still hold the old client, which may be the root cause of the new queries to fail.

If this is the case, would it make sense to export the initStores function 
```Javascript
function initStores(stores: Store<any, any>[]) {
  const client = getClient();
  if (!client) {
    throw new Error('Failed to initialize stores: MongoDB client not initialized');
  }

  for (const store of stores) {
    store.init(client);
  }
}
```
along with:

```Javascript
export const stores = getStores(combinedModules);
```

from file `packages\modelence\src\app\index.ts`
and then use it in db/client.ts after a reconnection, so that we can re-call initStores and also run recreateIndexes for the new DB client?

```Javascript
 if (mongodbUri) {
    for (const store of stores) {
      store.createIndexes();
    }
  }
  ```

Comment (omegascorp) [reply packages/modelence/src/db/client.ts]: Yes, but even if we rerun it here, it might be used anywhere in the code outside of the modelence, and since this reconnection issue is a rare case, people will not notice it until they face it.

There is a way to handle this without changing architecture - by using a proxy.

```
let realClient: MongoClient | null = null;
export const client = new Proxy({}, {
  get(_t, prop) {
    if (!realClient) {
      throw new Error("MongoClient proxy: not initialized yet (call init() first).");
    }
    const value = Reflect.get(realClient as any, prop, realClient);
    return typeof value === "function" ? value.bind(realClient) : value;
  },
  has(_t, prop) {
    return !!realClient && prop in realClient;
  },
  ownKeys() {
    return realClient ? Reflect.ownKeys(realClient) : [];
  },
  getOwnPropertyDescriptor(_t, prop) {
    return realClient
      ? Object.getOwnPropertyDescriptor(realClient, prop as any)
      : undefined;
  },
}) as unknown as MongoClient;
```

It means the client will always proxy all function calls to realClient, and we can replace realClient anytime without breaking any references.

@artahian since this is a core piece of logic, please check if you are ok with those changes

Comment (AtulYadav25) [reply packages/modelence/src/db/client.ts]: @omegascorp Should I proceed with adding this proxy client to the code?

Comment (omegascorp) [reply packages/modelence/src/db/client.ts]: Yeah, I think it will not work without this change anyway, my only worry is that it looks a bit complex

Comment (AtulYadav25) [reply packages/modelence/src/db/client.ts]: No worries, I'll make sure this works by testing after updates

Comment (AtulYadav25) [reply packages/modelence/src/db/client.ts]: Currently, if the Mongo URI is updated in `.env`, the app needs to restart, and on restart, it will reconnect to the new DB link. Are you asking me also to support dynamic DB URI changes at runtime (without restart)

If this is the case we can create and export new func
```Javascript
export async function switchDatabase(newUri) {
  if (client) {
    await client.close();
    console.log("Closed old MongoDB connection");
  }
  return connect(newUri);
}
```

This will allow updating the database without restart, but can you give scenarios why a user would need to switch it? (Very Rare IMO). So I think its better to keep it in env and allow restart to reconnect new database

omegascorp: The case I mentioned is when you reconnect (to the same URL), but there are still instances of MongoClient that are already disconnected.

AtulYadav25: I’ve updated the code in `db/client.ts` with a new method. While exploring the proxy-based approach, I noticed it would require removing the .init() usage from store.ts, since `.init()` currently looks like this (New Code):

```Javascript
  /** @internal */
  init(client: MongoClient) {
    if (this.collection && client === this.client) {
      throw new Error(`Collection ${this.name} is already initialized`);
    }

    this.client = client;
    this.collection = this.client.db().collection<this['_type']>(this.name);
  }
```

With the proxy method, I can import the client directly from `db/client.ts` and avoid calling .init(). 

After going through the codebase, I noticed that the root database handles come from `store.ts`. This uses this.client, which doesn’t serve a purpose under the proxy method and would have to be removed.

To keep changes minimal, I proceeded without the proxy method.

I tested this update and confirmed that it works as expected.
Here’s a short video demo of it in action:

https://github.com/user-attachments/assets/f8c5e45a-6d95-4337-b52f-1778e8fcdc30

AtulYadav25: If you’d prefer the proxy-based method instead, let me know and I can make those changes.

omegascorp:
@AtulYadav25 I'm sorry, but we decided to close this task and PR. The reason is that it's an edge case when the underlying database changes and the driver can't find the same user; in other cases, the MongoDB driver seems to self-heal. We believe that the default behavior is sufficient for now, and we don't want to make significant changes to address this case.

PR #68, opened 2025-09-01
Added several showcase examples

eduard-piliposyan:
Added several examples to showcase Modelence.


omegascorp (approved):
Looks good, I didn't manage to confirm that the image urls are correct in readme, in preview it shows broken images, but it might be just issue in diff

PR #70, opened 2025-09-03
WebSockets

omegascorp:
* Version updated to 0.6.0 since there is a new feature
* Support of the WebSockets
* Library used: socket.io
* Concept is simplified so you can directly have the concept of rooms.
* Create a channel on server via `new ServerChannel('roomName')`
* Create a channel on client via `new ClientChannel('roomName', onMessage)`
* Subscribe to a room via clientChannel.joinChannel(id)
* Broadcast messages via `serverChannel.broadcast(id, data)` - this will send a message to all clients subscribed to the room
* No proper docs yet
* New methods: use and all, they reflect the methods from express, the difference is that `all(/path, ...` will handle request targeting `/path`, on the other hand, `use(/path, ...` will handle all subpaths of `/path`

### DIFF
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.6.0-dev.10",
+  "version": "0.6.0",
   "description": "The Node.js Framework for Real-Time MongoDB Apps",
   "main": "dist/index.js",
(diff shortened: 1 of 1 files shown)


omegascorp:
@eduard-piliposyan @artahian I've fixed the architecture and moved the implementation of socket.io to a @modelence/socketio so we can add other providers in the future.

PR #76, opened 2025-09-18
Add support of custom MongoDB URI

omegascorp:
* Added _system.customMongodbUri
* Added _system.mongodbMode
* When _system.mongodbMode is 'custom' _system.customMongodbUri is used as mongodb URI


omegascorp:
Closing since we decided to support only one mongodb uri

PR #80, opened 2025-09-29
Data API Docs

omegascorp:
Added a page with Data API docs

### DIFF
--- a/docs/web/data-api.mdx
+++ b/docs/web/data-api.mdx
@@ -49,7 +49,7 @@ Your Data API will be available at `http://localhost:3000`
 
 ### DB Access
 
-The MongoDB URL for Data API can be configured via the dataApi.mongodbUri setting in Application → Configuration at https://cloud.modelence.com/.
+The MongoDB URL for Data API can be configured via the `dataApi.mongodbUri` setting in Application → Configuration at https://cloud.modelence.com/.
 
 <Note>
 The Modelence framework uses the `_system.mongodbUri` configuration to connect to MongoDB. You can configure two different MongoDB instances to separate Modelence-specific collections from those accessible via Data API. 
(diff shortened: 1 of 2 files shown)


docs/web/data-api.mdx
```diff
@@ -0,0 +1,553 @@
...
+```bash
+npx create-modelence-app@latest ai-chat --template data-api
+```
+
+### 2. Connect to Modelenece Cloud
+
+1. Open [cloud.modelence.com](https://cloud.modelence.com/) create
+2. Create a new application and a local environment
+3. Click on Setting -> Set up
+4. Copy the command that connects to remote DB, it should looks like the following:
```
eduard-piliposyan: During the setup steps there is no way to select remote DB. Do we need this step here?

docs/web/data-api.mdx
```diff
@@ -0,0 +1,553 @@
...
+Your Data API will be available at `http://localhost:3000`
+
+## Core Components
+
+### Authentication
+
+The Data API supports two authentication methods:
+
+#### 1. Direct API Key Authentication
+
```
eduard-piliposyan, line 63: Somewhere in this steps we should mention also that they need to configure `dataApi.mongodbUri` field.

docs/web/data-api.mdx
```diff
@@ -51,6 +43,14 @@ Your Data API will be available at `http://localhost:3000`
 
 ## Core Components
 
+### DB Access
+
+The MongoDB URL for Data API can be configured via the dataApi.mongodbUri setting in Application → Configuration at https://cloud.modelence.com/.
```
eduard-piliposyan: Probably better to have `dataApi.mongodbUri` (in the ` brackets) to show better in markdown.

omegascorp: Thank you, done

PR #81, opened 2025-10-01
Vite allowed hosts

artahian:
With these changes now we can run projects in dev mode under custom subdomains (needed for sandbox environments like tenant-123.sandbox.modelence.app).


omegascorp (approved):

omegascorp:
@artahian this should be handled by the order change in the vite server, sould we close this?

PR #82, opened 2025-10-01
Merge Vite config with user Vite config

omegascorp:


packages/modelence/src/viteServer.ts
```diff
@@ -92,33 +109,24 @@ async function getConfig() {
...
     resolve: {
       alias: {
         '@': path.resolve(appDir, 'src').replace(/\\/g, '/')
       }
     },
-    publicDir: userConfig.publicDir,
-  };
+  });
+
+  return safelyMergeConfig(baseConfig, userConfig);
```
artahian, line 129: @omegascorp I wonder if it's easier to have an explicit merge for each property above using the spread operator instead of a generic merge function? It's just really hard to understand how the merge works here, looking at the code, and I'm afraid we may need different logic for different properties.

Since the framework itself uses only a few, perhaps we can manually handle each one above? Also, I wonder if we should always give the user vite config higher priority so the user can always override anything they want? In the worst case they'll break things with invalid configs, but the good part is that we won't be locking them into anything (I've had this problem just yesterday where I had to make a change in framework code since I couldn't set the server vite config from the project itself).

omegascorp: It actually gives priority to the user config. The only thing this function does is remove duplicate plugins, as they often break the build. It also prioritizes user plugins over the Modelence default config plugins.

omegascorp: Note: actual merge is done by `mergeConfig`, the code I've added here works only with `mergedConfig.plugins`

artahian (approved):

PR #85, opened 2025-10-02
MongoDB Search Indexes

omegascorp:
* Allow adding search indexes
* Recreate indexes with names if they are changed

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.6.2",
+  "version": "0.6.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.6.2",
+      "version": "0.6.4",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@socket.io/mongo-adapter": "^0.4.0",
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.6.3",
+  "version": "0.6.4",
   "description": "The Node.js Framework for Real-Time MongoDB Apps",
   "main": "dist/index.js",
   "types": "dist/global.d.ts",


docs/web/docs.json
```diff
@@ -26,6 +26,12 @@
             "pages": [
               "tutorial"
             ]
+          },
+          {
+            "group": "Core Concepts",
```
artahian: @omegascorp isn't this supposed to be under API Reference? We already have Store and other things there:
https://docs.modelence.com/api-reference/modelence/server/classes/Store

The main distinction is that this main page is for guides and tutorials, while the API reference is just a full list of everything available.

docs/web/stores.mdx
```diff
@@ -0,0 +1,409 @@
+---
+title: 'Stores'
```
artahian, line 2: Same comment here as above - can we somehow consolidate both Store pages? I understand one is API reference and this one is more like a tutorial, but I feel like we're duplicating everything and a lot of these examples could be just Typedoc comments for each Store method that will automatically be displayed on API reference.

packages/modelence/src/data/store.ts
```diff
@@ -64,6 +66,7 @@ export class Store<
   private readonly schema: TSchema;
   private readonly methods?: TMethods;
   private readonly indexes: IndexDescription[];
+  private readonly searchIndexes: SearchIndexDescription[];
```
artahian, line 69: @omegascorp why do we need search indexes? Isn't this for full text search, as opposed to vector indexes, which is what we need?

Also, regardless of this question, I wanted to share something that I wrote a while ago in case it's helpful, since it's very related:
https://www.notion.so/AI-18d898b951c580d48d9fd647619e9234#18d898b951c580d48d9fd647619e9234

artahian (commented):
I have two high level questions:
- Documentation: looks like we're duplicating some things with API reference - can we keep the Store documentation in the Typedoc and have it still be autogenerated? I feel like currently it's just a list of what's possible, which is more of an "API reference", rather than a specific use case / showcase, which is more like a "tutorial" or "guide" for the home page
- For the indexes, I'm a bit confused on why we have search indexes instead of vector indexes here? Also, for either one, can we just make it a part of the same existing indexes array instead of introducing another property or is there a good reason to separate? I see that search indexes have a separate method in MongoDB's driver, but I believe vector indexes were just part of normal indexes?

docs/web/docs.json
omegascorp: The AI reference is auto-generated and somewhat complex to read, in my opinion. I want to generate more flexible examples. I will make it sorter so it's less like API reference

packages/modelence/src/data/store.ts
omegascorp: @artahian I think it looks simpler when they are separated. They have a different structure, and search indexes are only applicable to the Atlas.

artahian (approved):
Approving with the same comment as the related PR - as a next step let's add a helper that will make it easier for the user to specify the index

docs/web/stores.mdx
omegascorp: Let's improve on these things later; for now, I just want to add a page for each feature that we have. The goal of the features page is to give an understanding of how the feature works, goal of the API Reference is to provide a comprehensive documentation. They might overlap, but that should be fine. It's better if people read it multiple times than if they don't read it at all. Also, note that LLMS will be learning the docs - the duplication is great for them.

PR #91, opened 2025-10-03
fix: add entryPoints and tsconfig to typedoc.json for @modelence/next

lightning-sagar:
### Changes:

* Added the `entryPoints` and `tsconfig` fields to the `typedoc.json` configuration for the `@modelence/next` package
* Fixes the issue where Typedoc wasn't generating function docs, enabling auto-generation when running npm run docs
---

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -20,7 +20,7 @@
   "license": "ISC",
   "devDependencies": {
     "@types/express": "^5.0.3",
-    "modelence": "*",
+    "modelence": "^0.6.2",
     "next": "^15.4.6",
     "tsup": "^8.3.6",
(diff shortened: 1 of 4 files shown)


packages/next/package.json
```diff
@@ -20,7 +20,7 @@
   "license": "ISC",
   "devDependencies": {
     "@types/express": "^5.0.3",
-    "modelence": "*",
+    "modelence": "^0.6.2",
```
artahian: Can you revert this change? This will likely create problems with old versions of Modelence (we switched to `*` because of these issues originally)

artahian: cc @lightning-sagar

artahian (approved):

PR #93, opened 2025-10-06
Create Authentication page in Features section

omegascorp:
* Created Authentication page in Features section


eduard-piliposyan (approved):
All looks good to me. But there is nothing about social signup/login (e.g., Google). Do you think we should add information about them here as well, or you plan to create a separate page for that?

omegascorp:
@eduard-piliposyan updated, merging

PR #96, opened 2025-10-11
Add methods for vector search

omegascorp:


packages/modelence/src/data/store.ts
```diff
@@ -432,4 +432,109 @@ export class Store<
...
+    field: string,
+    embedding: number[],
+    numCandidates?: number;
+    limit?: number;
+    projection?: Document;
+  }) {
+    return this.aggregate([
+      {
+        $vectorSearch: {
+          index: field + 'VectorSearch',
```
artahian: Can we also make the index optionally customizable in case the user wants a custom index name?

packages/modelence/src/data/store.ts
```diff
@@ -432,4 +432,109 @@ export class Store<
...
+    dimensions,
+    similarity = 'cosine',
+  }: {
+    field: string;
+    dimensions: number;
+    similarity?: 'cosine' | 'euclidean' | 'dotProduct';
+  }) {
+    return {
+      type: 'vectorSearch',
+      name: field + 'VectorSearch',
```
artahian: Same comment as above - can we allow the user to optionally customize this?

omegascorp: Done, merging

PR #99, opened 2025-10-11
Add GitHub support

omegascorp:
* Add GitHub auth
* Fix the case when auth is enabled after the deployment

### DIFF
--- a/packages/auth-ui/package-lock.json
+++ b/packages/auth-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelence/auth-ui",
-  "version": "0.1.17-dev.0",
+  "version": "0.1.17",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelence/auth-ui",
-      "version": "0.1.17-dev.0",
+      "version": "0.1.17",
       "license": "ISC",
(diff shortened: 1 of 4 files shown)


packages/modelence/src/auth/providers/oauth-common.ts
```diff
@@ -0,0 +1,143 @@
...
+          id: userData.id,
+        },
+      },
+    });
+
+    await authenticateUser(res, newUser.insertedId);
+
+    const userDocument = await usersCollection.findOne(
+      { _id: newUser.insertedId },
+      { readPreference: "primary" }
```
artahian: I know this was pre-existing, but isn't this already the default everywhere?

omegascorp: The reads can go to replicates, `{ readPreference: "primary" }` forces to read from primary, we need it only in the cases when we read the document immidiately after inserting

artahian: Wait, isn't the default read preference always use the primary? We already make this assumption everywhere, so if that wasn't the case that would cause other issues in edge cases. If not already, we should just make this the default behavior (and only use secondary reads when explicitly specified otherwise).

PR #107, opened 2025-10-24
Added admin role support

eduard-piliposyan:


docs/web/admin-roles.mdx
```diff
@@ -0,0 +1,76 @@
+# Admin Roles
```
omegascorp: I think it's better to call the page `Roles` and make it more generic. Can we design it so it's easy to define custom roles? For example we can check isRole('admin') so it's easy to make things like isRole('moderator') or isRole('client')

artahian: Agreed - I wouldn't make this about admin, except maybe in some of the examples.

packages/modelence/src/auth/role.ts
```diff
@@ -51,3 +51,7 @@ export function hasPermission(roles: Role[], permission: Permission) {
 
   return false;
 }
+
+export function isAdmin(user: { roles?: string[] } | null): boolean {
+  return user?.roles?.includes('admin') ?? false;
+}
```
artahian: Aside from @omegascorp mentioned about using `hasRole` instead of `isAdmin`, I think we should also make this a method on the User data model, so you can call `user.hasRole('admin')` or `user.requireRole('admin')` (the second version will throw an exception so you can simply call `await user.requireRole('admin')` instead of doing checks every time.

But I think you also need to change the UserInfo type in the context we're passing to method calls, and pass the real user data model instead of a raw object.

packages/modelence/src/client/session.ts
```diff
@@ -81,3 +83,12 @@ export function useSession() {
...
   return { user };
 }
+
+export function isAdmin(user: User | null): boolean {
+  return user?.roles?.includes('admin') ?? false;
+}
+
+export function useIsAdmin(): boolean {
+  const user = useSessionStore((state) => state.user);
+  return isAdmin(user);
```
artahian: Same for the frontend - can we add a `hasRole` method on the user object instead of adding yet another `useIsAdmin` hook? In general, I think we need to avoid adding a separate React hook for specific checks like this, especially if the user itself already has a session hook that includes everything we need.

docs/web/roles.mdx
```diff
@@ -0,0 +1,92 @@
...
+      throw new Error('Admin access required');
+    }
+    // Admin-only logic here
+  }
+});
+
+// Or use requireRole for cleaner code
+createMutation('adminAction', {
+  handler: async (args, { user }) => {
+    user?.requireRole('admin'); // Throws error if not admin
```
artahian, line 22: @eduard-piliposyan btw this `requireRole` example has a security issue - it passes all checks for unauthenticated users. We still need an `if (!user) { throw new Error('Unauthenticated'); }` kind of a check before.

PR #110, opened 2025-10-27
Store extend

omegascorp:
* Add extend method to Store

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.6.12",
+  "version": "0.6.13",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.6.12",
+      "version": "0.6.13",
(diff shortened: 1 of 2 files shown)


packages/modelence/src/data/store.ts
```diff
@@ -109,6 +128,86 @@ export class Store<
...
+      schema: extendedSchema,
+      methods: combinedMethods as unknown as CombinedMethods | undefined,
+      indexes: extendedIndexes,
+      searchIndexes: extendedSearchIndexes,
+    });
+
+    // If this store is already initialized, initialize the extended store too
+    if (this.client) {
+      extendedStore.init(this.client);
+    }
```
artahian, line 212: @omegascorp shouldn't we just throw an error if the store is already initialized? We create indexes on startup, and there is a specific time & place for it when the app starts, so if you've missed that we probably shouldn't allow modifying anything after that, since that means the user is doing something wrong. If you extend an existing Store after things have been initialized, it means you are already creating something too late, after `startApp` has been called, and your extended version won't be properly provisioned and its indexes won't be created.

omegascorp: @artahian updated

PR #112, opened 2025-10-27
Add type validation to insertOne and insertMany

omegascorp:


artahian (commented):
@omegascorp `this['_type']` is already defined as `InferDocumentType<TSchema>` - can you share more context on this change, does `this['_type']` not work as expected?

omegascorp:
@artahian, no, for some reason it does not work, try writing dbUsers.insertOne({ - it does not give any suggests.

artahian (approved):

PR #113, opened 2025-10-29
Track performance of routes

omegascorp:
* Track performance of routes

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.6.14",
+  "version": "0.6.15",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.6.14",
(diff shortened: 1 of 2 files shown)


packages/modelence/src/routes/handler.ts
```diff
@@ -15,11 +16,19 @@ export function createRouteHandler(handler: RouteHandler) {
       try {
         const { session, user } = await authenticate(authToken);
         context = { session, user };
-      } catch (error) {
+      } catch {
         // If authentication fails, context remains null
       }
     }
 
+    const transaction = startTransaction('route', `route:${method.toLowerCase()}:${path}`, {
```
artahian, line 24: Is it scalable to make `path` a part of the name given that it can have many different variations (even for the same route)? I think there are some indexing limits on how many unique values you can use.

omegascorp: @artahian path is what we give to express. For example: user/:id (not user/123). Method + Path uniquely define the route.

PR #115, opened 2025-11-04
Improve Cron Job locks

omegascorp:
* Add heartbeat to locks
* Acquire a lock before executing the cron jobs (it's a single global lock for all cron jobs)
* Lock can be acquired only when the heartbeat times out
* Use random container IDs instead of MODELENCE_CONTAINER_ID
* Renamed site module to system
* Added MODELENCE_LOG_LEVEL
* Renamed MODELENCE_SITE_ENV to MODELENCE_ENV (while it's not used anywhere, I think it's a better choice)

### DIFF
--- a/packages/modelence/src/cron/jobs.ts
+++ b/packages/modelence/src/cron/jobs.ts
@@ -109,12 +109,12 @@ async function tickCronJobs() {
     // TODO: limit the number of jobs running concurrently
 
     if (state.scheduledRunTs && state.scheduledRunTs <= now) {
-      await startCronJob(job);
+      await runCronJob(job);
     }
   });
 }
 
-async function startCronJob(job: CronJob) {
+async function runCronJob(job: CronJob) {
   const { alias, params, handler, state } = job;
   state.isRunning = true;
   state.startTs = Date.now();
--- a/packages/modelence/src/lock/db.ts
+++ b/packages/modelence/src/lock/db.ts
@@ -4,8 +4,8 @@ import { Store } from '../data/store';
 /**
  * Database collection for storing distributed locks.
  *
- * Locks are used to ensure that only one instance of the application can acquire
- * a lock for a specific container at a time.
+ * Locks are used to ensure that only one instance of the application can perform
+ * a specific action at a time, such as running cron jobs or migrations
  */
 export const locksCollection = new Store('_modelenceLocks', {
   schema: {
--- a/packages/modelence/src/lock/helpers.ts
+++ b/packages/modelence/src/lock/helpers.ts
@@ -57,7 +57,7 @@ export async function acquireLock(
   } = {}
 ): Promise<boolean> {
   const now = Date.now();
-  if (lockCache[resource] && now - lockCache[resource].expiresAt < 0) {
+  if (lockCache[resource] && now < lockCache[resource].expiresAt) {
     return lockCache[resource].value;
   }
 


packages/modelence/src/cron/jobs.ts
```diff
@@ -10,25 +11,34 @@ import { Store } from '../data/store';
...
+ * Interval at which lock heartbeats are updated in the database
  */
-const LOCK_TRANSFER_DELAY = time.seconds(10);
+const LOCK_HEARTBEAT_INTERVAL = time.seconds(5);
+
+/**
+ * Time after which a lock is considered stale if no heartbeat update was received
+ * Should be significantly larger than LOCK_HEARTBEAT_INTERVAL to account for network delays
+ */
+const LOCK_STALE_THRESHOLD = time.seconds(30);
```
artahian: 30s seems a bit long, can we do 15s or at least 20s instead? Even if the previous cron is still alive but is stuck doing some synchronous calculations, I think it's ok if one of the other cron jobs "steals" the lock.

packages/modelence/src/cron/jobs.ts
```diff
@@ -10,25 +11,34 @@ import { Store } from '../data/store';
...
   schema: {
     alias: schema.string(),
     lastStartDate: schema.date().optional(),
-    lock: schema
-      .object({
-        containerId: schema.string(),
-        acquireDate: schema.date(),
-      })
-      .optional(),
+    lock: schema.object({
```
artahian: Shouldn't this still be `.optional()`? Do we have a lock as soon as the cron job is inserted?

packages/modelence/src/cron/jobs.ts
```diff
@@ -75,97 +85,205 @@ export async function startCronJobs() {
...
 
-    const existingLockedRecord = await cronJobsCollection.findOne({
+    const now = new Date();
+    const staleThreshold = new Date(now.getTime() - LOCK_STALE_THRESHOLD);
+
+    // Find existing locks that are still active (not stale)
+    const activeLocks = await cronJobsCollection.fetch({
       ...aliasSelector,
       'lock.containerId': { $exists: true },
+      'lock.lastHeartbeat': { $gte: staleThreshold },
```
artahian: @omegascorp with this new heartbeat logic, I'm afraid it's too expensive to update a heartbeat in every single cron job document separately. There can be 100s of cron jobs, and updating 100s of documents every 5s is way too much.

Not sure if you can think of a better approach, but can we just store a separate collection for just cron locks, and update the heartbeat in the lock, then set a `lockId` on cron jobs?

It can still allow for multiple locks, for cases when you just deployed a new version with new cron jobs, that the old instance doesn't even know about. The new instance can create its own lock (i.e. the locks collection can always have 1-1 mapping with instances), and set it on new cron jobs that didn't have a previous lockId already. That way both instances can still do their own job while having only two heartbeat updates instead of hundreds.

Or, basically this new collection can be just "instances" instead of locks, while leaving the lock logic as it is. And the heartbeat can be at the instance level, not lock level. And the lock can store an `instance` or `container` ID. When one of the containers' heartbeat expires, one of the other containers can take over its locks and attempt to update their "containerId" to its own. The container that's successful in doing so before others (it should be a safe update like `update({ 'lock.containerId': expiredContainerId }, { $set: { 'lock.containerId': myContainerId } })` to account for race conditions) will claim these locks and continue running them.

We would also need a way for the other instance to recognize that it has been kicked out (in case it comes back after some time and wants to continue running jobs).

omegascorp: @artahian, what if we have only one instance running cron jobs? If an instance with a newer version is deployed, we can allow it to steal the lock

artahian: Sure, I think there is no point in keeping them split out on purpose. The new instance can take over and wait for a grace period to allow the old instance see the change and stop running jobs.

We can still store each instance in a collection, but only mark one as the cron job instance at a time. And this collection can be general - not cron specific, so it can be used for other things in the future as well.

omegascorp: Thinking around this, do we actually need to override the cron jobs? When the new version is deployed, the old one will likely die soon anyway

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,134 @@ export function defineCronJob(
...
-    cronJobsInterval = setInterval(tickCronJobs, time.seconds(1));
+  if (!lockAcquired) {
+    // There's another active instance holding the cron lock
+    if (!cronJobsInterval) {
+      console.log('Cron jobs are locked by another instance, will retry later');
+    }
+    return;
   }
+
+  console.log('Acquired cron lock, starting cron jobs');
```
omegascorp: @artahian, I'm thinking of creating an internal function for logging. For example, this log is useful for debugging, but useless for everyday work. I would prefer to show it only when logging level is debug

packages/modelence/src/cron/jobs.ts
artahian: Yeah I guess we don't need special logic like "If a new version exists then it should always take over the old version, even when both are alive"

packages/modelence/src/cron/jobs.ts
artahian: @omegascorp we already have log functions in the framework, my plan was already to expose it and let users use it as well.

https://github.com/modelence/modelence/blob/d4cddd720cd27afa0b3acc864c0d8ac010ed448c/packages/modelence/src/telemetry/index.ts#L10

We should just improve it to accept JSON logs as well, I really like to be able to send structured logs instead of just text (like `{ source: 'cron', event: 'lockAcquired', containerId: '...' }`)

packages/modelence/src/lock/db.ts
```diff
@@ -0,0 +1,36 @@
+import { schema } from '../data/types';
```
artahian, line 1: Can we call this file `db.ts` for consistency with other existing collection files?

packages/modelence/src/lock/collection.ts
```diff
@@ -0,0 +1,36 @@
...
+ *   heartbeatAt: new Date(),
+ * });
+ * ```
+ */
+export const locksCollection = new Store('_modelenceLocks', {
+  schema: {
+    type: schema.string(),
+    containerId: schema.string(),
+    acquiredAt: schema.date(),
+    heartbeatAt: schema.date(),
```
artahian: I don't think `heartbeatAt` is a valid name since if it ends with `At` the first part usually has to be an action. Maybe `lastHeartbeatAt` or just `lastHeartbeatDate`?

packages/modelence/src/lock/db.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @example
+ * ```typescript
+ * // Acquire a lock for a container
+ * const lock = await locksCollection.insertOne({
+ *   containerId: 'my-job',
+ *   acquiredAt: new Date(),
+ *   heartbeatAt: new Date(),
+ * });
+ * ```
+ */
```
artahian, line 9: Do we really need to document this collection? Shouldn't it be internal only?

I guess this is a higher level question, but I can't think of any other usages for container locks, and especially because this is container specific I'm not sure if it's worth generalizing it so much and giving it a "type".

Also, maybe we should call the collection `_modelenceContainerLocks` in case we need other locks in the future? There can be different types of locking mechanisms, but it usually is not all using the same pattern, so each one probably needs its own implementation.

packages/modelence/src/lock/helpers.ts
```diff
@@ -0,0 +1,180 @@
+import { randomBytes } from 'crypto';
+import { locksCollection } from './collection';
+
+/**
+ * Unique identifier for this application instance.
+ * Generated once per application instance to track which container owns which locks.
+ */
+const containerId = randomBytes(32).toString('base64url');
```
artahian: Since this code is container specific, would it make more sense to call this whole directory of files `instances` or `containers` instead of locks?

packages/modelence/src/lock/collection.ts
artahian: Btw, the heartbeat is also a container property rather than a lock concept, so I'm not sure I understand what a "lock heartbeat" means. Can we just move away from the whole "lock" concept, call this collection `_modelenceInstances` or `_modelenceContainers` and just mark one of the containers as the "cron instance"?

I feel like we need two separate collections, one for containers, with their IDs, heartbeat, status, etc, and another one with a single document that just stores whichever containerId is the "cron owner", so it can be atomically updated (essentially it is the "lock").

packages/modelence/src/lock/helpers.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Updates the heartbeat timestamp for a lock owned by this container.
+ *
+ * @param type - The type of lock to update
+ * @returns true if heartbeat was updated, false if lock wasn't owned by this container
+ */
+async function updateLockHeartbeat(type: string): Promise<boolean> {
+  const result = await locksCollection.updateOne(
```
artahian: Optional: can we keep a memory state whether we own the lock, and only attempt to update the heartbeat if we do? When an application scales to 10 or more containers, there will essentially be a lot of containers each trying to update their own heartbeat, even though only one really owns the lock, and the rest of the updates are going to be a waste and create extra load on the database.

It may still be ok without this, but if it's easy to check that could save a lot of unnecessary db update attempts.

packages/modelence/src/cron/jobs.ts
```diff
@@ -6,29 +6,23 @@ import { startTransaction, captureError } from '@/telemetry';
...
 const DEFAULT_TIMEOUT = time.minutes(1);
 
 /**
- * Each cron instance acquires locks for the jobs it runs. If there was a pre-existing lock,
- * the lock is transferred to the new instance, but there is a delay to give the previous instance
- * a chance to see the new lock and gracefully finish remaining jobs.
+ * Interval at which we retry acquiring the cron lock if we don't have it
  */
-const LOCK_TRANSFER_DELAY = time.seconds(10);
+const LOCK_RETRY_INTERVAL = time.seconds(30);
```
artahian: Isn't 30s too long? In the worst case it will be "lock expiration" + "retry interval", which is 30 + 30 = 60s. That's too long of an idle period for critical crons.

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,134 @@ export function defineCronJob(
...
+    return;
+  }
+
+  cronJobsStarted = true;
 
-    if (existingLockedRecord) {
-      await sleep(LOCK_TRANSFER_DELAY);
+  await tryAcquireLockAndStartJobs();
+
+  setInterval(async () => {
```
artahian: I don't think we need this second interval since we already have a 1s tick interval. We can use the existing interval to check if enough time has passed since our last attempt to acquire the lock and retry again if so - that's probably more reliable since there is only one interval and we control the exact order of everything happening instead of two parallel intervals.

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,134 @@ export function defineCronJob(
...
       }
-      return;
+      continue;
     }
 
     // TODO: limit the number of jobs running concurrently
 
     if (state.scheduledRunTs && state.scheduledRunTs <= now) {
-      await startCronJob(job);
+      const ownsLock = await verifyCronLock();
```
artahian: Why do we need to verify the cron job ownership in two places?

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,134 @@ export function defineCronJob(
...
+  if (!ownsLock && cronJobsInterval) {
+    console.log('Lost cron lock, stopping job execution');
+    clearInterval(cronJobsInterval);
+    cronJobsInterval = null;
+  }
 }
 
 async function tickCronJobs() {
-  // TODO: periodically check if the locks are still there
+  await verifyLockOwnershipAndStop();
```
artahian: `tickCronJobs` runs every second, so verifying the ownership on each tick is probably going to be too much, especially with multiple instances.

A few notes:
- The ownership expiration is much longer than 1s, so we can probably be smarter than just checking it every second
- Even with checking the ownership right before running anything, we are still at risk or race conditions, where the lock expires immediately after our check, and we're still running the cron job. 
- Can we instead skip the verification altogether if we are still in the safe zone after the last heartbeat? i.e. every time we confirm that our heartbeat was successfully updated in the database and we still owned the lock, we know that no one is supposed to kick the current container out for X seconds (X = lock expiration), so if we have a single interval and do everything in the right order we probably never really need lock verification? The container is always supposed to update its heartbeat before it even gets close to losing the lock, so the heartbeat itself can be used as verification.

So "verify ownership" can probably be replaced by "last heartbeat was less than Y seconds ago" where Y > heartbeat_delay but Y < lock_expiration

packages/modelence/src/lock/db.ts
omegascorp: I've renamed containerId to instanceId and allow to pass it as argument so it can be something else

packages/modelence/src/lock/helpers.ts
omegascorp: We can move instance / container id somewhere else, I've chaged it in a way you can pass a custom instanceId to acquire lock, I would prefer to keep it in the lock folder, I think it's more logical

packages/modelence/src/lock/helpers.ts
omegascorp: Heartbeat is removed

packages/modelence/src/cron/jobs.ts
omegascorp: Every container will be retrying to acquire the lock, I would prefer to keep it at least 30 seconds. It's 60 seconds if you have only 2 containers and one is dead. If you have 6 active containers it will retry every 5 seconds in average

packages/modelence/src/system/index.ts
```diff
@@ -0,0 +1,26 @@
+import { Module } from '../app/module';
+
+export default new Module('_system', {
```
artahian, line 3: @omegascorp I'm not sure how I feel about having a `_system` module. I think one Module definitely shouldn't be a prefix of another - it can lead to weird issues in the future. Maybe we can just make configs global and not force them into the Module structure?

packages/modelence/src/telemetry/index.ts
```diff
@@ -1,15 +1,36 @@
 import { getLogger, getApm } from '@/app/metrics';
 import { isTelemetryEnabled } from '@/app/state';
+import { getConfig } from '@/config/server';
+
+function getLogLevel() {
+  return getConfig('_system.log.level') as 'debug' | 'info' | 'error';
```
artahian, line 6: Out of the scope of this PR, but in the future we should find a way to derive these types correctly from the config schema.

packages/modelence/src/cron/jobs.ts
artahian: My concern is that it is likely that all containers will start at the same time and will be on the same schedule, still leaving the same gap. But if you want to keep it 30s, another option is to just try to acquire shortly when the last known date was expired instead of blindly trying.

But overall we can probably just leave this as it is since this is only happening during failover / deploys and 60s perhaps is not so bad in these cases.

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,104 @@ export function defineCronJob(
...
-    Object.values(cronJobs).forEach((job) => {
-      if (!job.state.scheduledRunTs) {
-        job.state.scheduledRunTs = now;
-      }
-    });
+/**
+ * Attempts to acquire the cron lock and start the cron job scheduler.
+ * If the lock is acquired, initializes job schedules and starts the tick interval.
+ */
+async function tryAcquireLockAndStartJobs() {
```
artahian: Minor, but this function seems to be asking to be split into two since it does two completely separate things and it's in the name.

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,104 @@ export function defineCronJob(
...
       }
-      return;
+      continue;
     }
 
     // TODO: limit the number of jobs running concurrently
 
     if (state.scheduledRunTs && state.scheduledRunTs <= now) {
-      await startCronJob(job);
+      const ownsLock = await acquireLock('cron');
```
artahian: Why do we need this check for each cron job? I think we just need to re-acquire the lock at the beginning of the tick function if it's time for the next "heartbeat", and if it's successful there are no other checks needed?

packages/modelence/src/cron/jobs.ts
```diff
@@ -66,106 +60,104 @@ export function defineCronJob(
...
 }
 
 async function tickCronJobs() {
-  // TODO: periodically check if the locks are still there
-
   const now = Date.now();
-  Object.values(cronJobs).forEach(async (job) => {
+  const jobs = Object.values(cronJobs);
+
+  for (const job of jobs) {
```
artahian: Looks like with this change you're making everything sequential and actually waiting for each job to complete? I don't think it was intended - cron jobs can be running for a long time, and you'll be blocking the whole tick function this way.

packages/modelence/src/system/index.ts
omegascorp: We don't have a proper way of making configs global, but this works for now, it's the easiest way to define configs without schema (it it's not something that will concern the users, so we can change this in the future if we define global configs)

packages/modelence/src/cron/jobs.ts
```diff
@@ -123,14 +85,18 @@ export async function startCronJobs() {
...
-function sleep(ms: number) {
-  return new Promise((resolve) => setTimeout(resolve, ms));
-}
-
 async function tickCronJobs() {
-  // TODO: periodically check if the locks are still there
-
   const now = Date.now();
+
+  const ownsLock = await acquireLock('cron', {
```
omegascorp, line 90: The failed lock is saved for 30 seconds, so it will not do another DB request before that.

artahian: Great, I like the caching logic - so the code that's using this doesn't have to worry about delays / frequency at all and can simply call it and let `acquireLock` take care of caching.

packages/modelence/src/cron/jobs.ts
```diff
@@ -174,6 +127,20 @@ async function startCronJob(job: CronJob) {
       },
     }
   );
+
+  const transaction = startTransaction('cron', `cron:${alias}`);
+  // TODO: enforce job timeout
+  try {
+    await handler();
```
artahian, line 134: Btw, since you've switched to await-in on the handler and this `startCronJob` function now waits until the job is done, might be better to rename it to `runCronJob` (previously the point of "start" was that it starts it but doesn't wait until completion). This can help to not accidentally await on `startCronJob` in the future thinking that it will return quickly after having started the job.

packages/modelence/src/lock/db.ts
```diff
@@ -0,0 +1,28 @@
+import { schema } from '../data/types';
+import { Store } from '../data/store';
+
+/**
+ * Database collection for storing distributed locks.
+ *
+ * Locks are used to ensure that only one instance of the application can acquire
+ * a lock for a specific container at a time.
```
artahian:
```suggestion
 * Locks are used to ensure that only one instance of the application can perform
 * a specific action at a time, such as running cron jobs or migrations
```

packages/modelence/src/lock/helpers.ts
```diff
@@ -0,0 +1,148 @@
...
+    instanceId = INSTANCE_ID,
+  }: {
+    lockDuration?: number;
+    successfulLockCacheDuration?: number;
+    failedLockCacheDuration?: number;
+    instanceId?: string;
+  } = {}
+): Promise<boolean> {
+  const now = Date.now();
+  if (lockCache[resource] && now - lockCache[resource].expiresAt < 0) {
```
artahian: Why not do this instead?
```suggestion
  if (lockCache[resource] && now < lockCache[resource].expiresAt) {
```

packages/modelence/src/lock/helpers.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @param resource - The resource to release the lock for
+ * @param options.instanceId - The unique identifier for this application instance
+ * @returns true if lock was released, false if lock wasn't owned by this container
+ */
+export async function releaseLock(
+  resource: string,
+  {
+    instanceId = INSTANCE_ID,
+  }: {
+    instanceId?: string;
```
artahian, line 137: Shouldn't this be a non-optional parameter? Otherwise if it's undefined, the `deleteOne` can have unintended consequence and will delete whatever lock there is, even if it's not owned by this instance.

omegascorp: The instanceId has a default value, so if it's not provided it will use the current instance id

omegascorp:
@artahian suggestions are applied, merging

packages/modelence/src/lock/helpers.ts
artahian: Sure, functionally I think it's good, but shouldn't the type be `instanceId: string` then without `?` ?

PR #120, opened 2025-11-10
Async migrations

omegascorp:
* Run migrations async
* Use migrations lock to run it on a single instance
* Release the lock once migrations are completed
* Write migration status and output to db

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.6.17",
+  "version": "0.6.18",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.6.17",
+      "version": "0.6.18",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@socket.io/mongo-adapter": "^0.4.0",
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.6.17",
+  "version": "0.6.18",
   "description": "The Node.js Framework for Real-Time MongoDB Apps",
   "main": "dist/index.js",
   "types": "dist/global.d.ts",


packages/modelence/src/migration/index.ts
```diff
@@ -1,35 +1,97 @@
...
+    });
+    try {
+      const output = await handler();
+      await dbMigrations.upsertOne(
+        {
+          version,
+        },
+        {
+          $set: {
+            status: 'completed',
```
eduard-piliposyan: @omegascorp Migrations can also have a "description" field, which is very helpful to keep in the DB, alongside versions.

packages/modelence/src/migration/index.ts
```diff
@@ -1,35 +1,97 @@
...
+    });
+    return;
+  }
+
   const versions = migrations.map(({ version }) => version);
 
-  const existingVersions = await dbMigrations.fetch({ version: { $in: versions } });
+  const existingVersions = await dbMigrations.fetch({
+    version: { $in: versions },
+    status: { $ne: 'failed' },
```
eduard-piliposyan: Does this mean that if a migration script fails, it will try to run it repeatedly every time it starts?

omegascorp: Yes, let me handle this differently. I think Aram was suggesting to mark them as failed and allow to rerun from the cloud.modelence.com

omegascorp: @eduard-piliposyan, the new condition is removed - the failed migrations will be just marked as failed.

packages/modelence/src/migration/index.ts
omegascorp: @eduard-piliposyan done

packages/modelence/src/app/index.ts
```diff
@@ -125,7 +125,7 @@ export async function startApp({
   }
 
   if (isCronEnabled) {
-    await runMigrations(migrations);
+    startMigrations(migrations);
```
artahian, line 128: @omegascorp I think we can get rid of the isCronEnabled check here and everywhere else since the locks should handle everything now.

The only reason for the cron enabled flag was to set it on a single instance only, but that was already not the case on Heroku anyway.

packages/modelence/src/migration/index.ts
```diff
@@ -1,35 +1,98 @@
...
-    console.log(`Running migration v${version}: ${description}`);
-    // TODO: adjust to handle multiple containers and race conditions
-    await dbMigrations.insertOne({ version, appliedAt: new Date() });
-    await handler();
-    console.log(`Migration v${version} complete`);
+    logInfo(`Running migration v${version}: ${description}`, {
+      source: 'migrations',
+    });
+    try {
+      const output = await handler();
```
artahian, line 46: @omegascorp I think the danger in swapping the order here is that if a migration script is getting stuck and never resolving with a success or fail, we will keep trying to run it all the time?

packages/modelence/src/migration/index.ts
```diff
@@ -1,35 +1,98 @@
...
-  handler: () => Promise<void>;
+  handler: () => Promise<string | void>;
 };
 
 export async function runMigrations(migrations: MigrationScript[]) {
   if (migrations.length === 0) {
     return;
   }
 
+  const hasLock = await acquireLock('migrations');
```
artahian, line 17: What happens if the current owner of the lock is a different version of the app that doesn't have all the up to date migrations?

More of a theoretical question since I find it hard to imagine how that would happen, but still would be good to have some guardrails for that, like attempting to run migrations periodically while there are pending scripts and ending that cycle only when there are no more pending scripts.

PR #121, opened 2025-11-11
Add deleteUsers and softDeleteUsers methods

omegascorp:
* Add deleteUser and disableUser methods
* Add a status field to the user

### DIFF
--- a/packages/modelence/src/auth/db.ts
+++ b/packages/modelence/src/auth/db.ts
@@ -26,7 +26,9 @@ export const usersCollection = new Store('_modelenceUsers', {
         })
       )
       .optional(),
+    status: schema.enum(['active', 'disabled', 'deleted']).optional(),
     createdAt: schema.date(),
+    disabledAt: schema.date().optional(),
     deletedAt: schema.date().optional(),
     roles: schema.array(schema.string()).optional(),
     authMethods: schema.object({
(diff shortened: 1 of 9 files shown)


packages/modelence/src/auth/db.ts
```diff
@@ -27,6 +27,7 @@ export const usersCollection = new Store('_modelenceUsers', {
       )
       .optional(),
     createdAt: schema.date(),
+    deletedAt: schema.date().optional(),
```
artahian, line 32: I think we need a status field as well because the existence of a field can't be efficiently queried. Also we may need the status for other states like "disabled" in the future

packages/modelence/src/server.ts
```diff
@@ -15,6 +15,7 @@ export { createQuery } from './methods';
 // Auth
 export { usersCollection as dbUsers } from './auth/db';
 export type { UserInfo } from './auth/types';
+export { deleteUser, softDeleteUser } from './auth/deleteUser';
```
artahian: Can we call it disableUser instead of softDeleteUser and have a separate disabled status? Later the consumer of this code can choose their custom logic, like disable immediately but then delete after 48 hours for example.

Our current delete is also technically soft delete, so it can be confusing as a name this way.

omegascorp: I was checking the best practices, and it seems it's better to leave anonymized records of the users since it might be required for auditing logs

artahian: Sure, we definitely need to keep the document itself, but what I mean in the comment is that the current softDeleteUser function seems confusing because it sets "deletedAt" while still keeping the handle and emails. I think we can simply call it "disableUser" instead.

omegascorp: @artahian done, the status is added

PR #124, opened 2025-11-13
Convert migration output to string

omegascorp:


packages/modelence/src/migration/index.ts
```diff
@@ -53,7 +53,7 @@ export async function runMigrations(migrations: MigrationScript[]) {
             version,
             status: 'completed',
             description,
-            output: output || '',
+            output: (output || '').toString(),
```
artahian: Not a blocker for now, but in the future I would also trim this since MongoDB has a 16mb hard limit and if the user returns a very large data the insert will fail.

artahian (approved):

PR #138, opened 2025-12-16
Add Core Concepts section to docs

omegascorp:


docs/web/core-concepts/modules.mdx
```diff
@@ -0,0 +1,419 @@
...
+
+  const handleSubmit = () => {
+    createTodo({ title: 'New todo', dueDate: new Date() });
+  };
+
+  return <button onClick={handleSubmit}>Create Todo</button>;
+}
+```
+
+## Stores
```
eduard-piliposyan: Maybe it's better to talk about Stores first and then about Mutations and Queries as all examples use dbTodos, which is defined only here in Stores.

eduard-piliposyan (commented):
Looks great. I have one suggestion. Approving.

docs/web/core-concepts/modules.mdx
omegascorp: Done

artahian:
@omegascorp general question - how are the Core Concepts different from Features? I understand why Project Structure is there, but Modules and Configurations don't seem any different to me than Stores or Telemetry.

PR #139, opened 2025-12-17
MOD-157: Add body options to routes

omegascorp:
* Added the following three options: json, urlencoded, and raw
* To disable JSON parsing and enable the raw parser, set the following
```
{
  json: false,
  raw: true,
}
```


eduard-piliposyan (approved):
Looks simple and good to me. We can make it even more generic by allowing users to define their own middleware.
But for now, this looks fine to me!

artahian:
@omegascorp the part where you have to pass both json:false and raw:true seems somewhat confusing - is there any other way to structure this? Can both be true?

PR #143, opened 2025-12-25
Fix WebSocket channel authorization bypass

agarwal-tanmay-work:
Removes an unconditional socket.join(channelName) call that allowed clients to
join channels even when authorization checks failed.

Channel joins now occur only after access validation succeeds, preventing
unauthorized subscription to private or sensitive channels.


agarwal-tanmay-work:
Thank you very much :)

PR #145, opened 2025-12-25
Read default config when remote server is not enabled

omegascorp:
* The system was setting the default value only when the remote backend was enabled
* Added fallback option to read the default value from the schema

### DIFF
--- a/packages/modelence/src/config/server.ts
+++ b/packages/modelence/src/config/server.ts
@@ -41,7 +41,7 @@ let isInitialized = false;
  * ```
  */
 export function getConfig(key: ConfigKey) {
-  return config[key]?.value;
+  return config[key]?.value ?? configSchema[key]?.default;
 }
 
 export function getPublicConfigs() {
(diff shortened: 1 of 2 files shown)


packages/modelence/src/config/server.ts
```diff
@@ -41,7 +41,7 @@ let isInitialized = false;
  * ```
  */
 export function getConfig(key: ConfigKey) {
-  return config[key]?.value;
+  return config[key]?.value ?? configSchema[key]?.default;
 }
```
eduard-piliposyan, line 45: @omegascorp should we do similar change for the client side as well ? in "/src/config/client.ts"

omegascorp: It's already handled for the client configs https://github.com/modelence/modelence/blob/6a57d0b73389c5d034efd2033f569de6c90366a1/packages/modelence/src/config/server.ts#L63

eduard-piliposyan: good then I guess we can merge this

PR #148, opened 2025-12-30
Email verification hook

omegascorp:
Added onAfterEmailVerification hook
Added onEmailVerificationError hook
Exported AuthSuccessProps type
Exported AuthErrorProps type

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.8.0",
+      "version": "0.9.0",
       "license": "SEE LICENSE IN LICENSE",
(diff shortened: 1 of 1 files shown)


packages/modelence/src/auth/verification.ts
```diff
@@ -62,9 +64,37 @@ export async function handleVerifyEmail(params: RouteParams): Promise<RouteRespo
 
     // Delete the used token
     await emailVerificationTokensCollection.deleteOne({ _id: tokenDoc._id });
+
+    const authConfig = getAuthConfig();
+    authConfig.onAfterEmailVerification?.({
+      provider: 'email',
+      user: (await usersCollection.findOne({ 'emails.address': tokenDoc?.email })) as User,
+      session: null,
+      connectionInfo: {
```
eduard-piliposyan, line 73: Why do we need this info in onAfterEmailverification?

eduard-piliposyan (approved):

packages/modelence/src/auth/verification.ts
omegascorp: It's used for analytics

PR #163, opened 2026-02-02
feat: add _modelence_ prefix to indexes

omegascorp:
* Add _modelence_ prefix to indexes
* Drop indexes with _modelence_ prefix that does not exist in the code
* Drop conflicting indexes

---

> [!NOTE]
> **Medium Risk**
> Touches index creation/deletion behavior and may drop/replace existing DB indexes during startup based on reconciliation logic, which can impact query performance and production migrations if misconfigured.
> 
> **Overview**
> **MongoDB index management now namespaces framework-created indexes** by automatically normalizing all `Store` index names to use an `_modelence_` prefix (including auto-generated names).
> 
> `Store.createIndexes()` is reworked to *reconcile* desired indexes against live DB metadata: it lists existing indexes, drops orphaned `_modelence_` indexes no longer defined in code, and replaces conflicting indexes (by name/options or by identical key under a different name), while preserving non-conflicting user-created indexes. It also hardens behavior for missing collections (Mongo error code `26`) and ignores index-not-found races during drops (code `27`), while keeping the existing retry-on-duplicate behavior for Atlas Search indexes (code `68`).
> 
> Bumps `modelence` package version to `0.12.0` and expands tests to cover the new prefixing and reconciliation scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1c0051dbfa8f77cabb5744f5a337280416fbc37. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.11.2",
+  "version": "0.12.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.11.2",
+      "version": "0.12.0",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@socket.io/mongo-adapter": "^0.4.0",
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.11.2",
+  "version": "0.12.0",
   "description": "The Node.js Framework for Real-Time MongoDB Apps",
   "main": "dist/index.js",
   "types": "dist/global.d.ts",


packages/modelence/src/data/store.ts
```diff
@@ -349,14 +427,38 @@ export class Store<
...
       for (const index of this.indexes) {
         try {
-          await this.requireCollection().createIndexes([index]);
+          await collection.createIndexes([index]);
         } catch (error) {
-          if (error instanceof MongoError && error.code === 86 && index.name) {
-            await this.requireCollection().dropIndex(index.name);
-            await this.requireCollection().createIndexes([index]);
+          if (error instanceof MongoError && (error.code === 86 || error.code === 85)) {
+            const indexName = extractIndexNameFromError(error);
```
eduard-piliposyan: Don't we have `index.name` here? Extracting index names from the error message looks risky to me, as the MongoDB error message can change in the future.

packages/modelence/src/data/store.ts
```diff
@@ -80,6 +80,83 @@ type EnhancedFilterOperators<T> = Omit<FilterOperators<T>, '$in' | '$nin'> & {
...
+    }
+    throw error;
+  }
+};
+
+/**
+ * Generates an auto-generated index name from the index keys
+ * Mimics MongoDB's default naming: field1_direction1_field2_direction2
+ */
+const generateAutoIndexName = (key: Document): string => {
```
eduard-piliposyan, line 189: @omegascorp, it seems we don't handle unique indexes here.
If an index is unique, it will have the same name as a non-unique index, right?
In that case, our main issue won't be resolved, as it occurs when switching from non-unique to unique.

packages/modelence/src/data/store.ts
```diff
@@ -80,39 +82,129 @@ type EnhancedFilterOperators<T> = Omit<FilterOperators<T>, '$in' | '$nin'> & {
...
+  key?: Document;
+  name?: string;
+};
 
-    // Pattern for code 86: "requested index: ... name: "indexName""
-    const requestedMatch = message.match(/requested index:.*?name:\s*"([^"]+)"/i);
-    if (requestedMatch?.[1]) {
-      return requestedMatch[1];
+const COMPARABLE_INDEX_OPTION_FIELDS = [
+  'background',
```
artahian, line 91: Just FYI, but I think maybe `background` is ok to skip from this list since it's still the same index, whether created in the background or not.

artahian (approved):
I was hoping this would be much simpler 😅 
Approving, but didn't go too deep into each part of the logic since it's a lot. I hope the test will catch if there are issues.

PR #172, opened 2026-02-04
fix(auth): handle GitHub OAuth when user email is private

### LINKED ISSUE #170: Bug: GitHub OAuth login fails when user email is private
When signing in using GitHub OAuth, authentication fails if the GitHub account does not have a public email configured.

By default, many GitHub accounts have “Keep my email address private” enabled.
In this case, the GitHub `/user` API returns `email: null`, even when requesting the `user:email` scope.

The current authentication flow assumes `githubUser.email` is always present, which causes login to fail for users with private emails.

### Steps to Reproduce
1. Use a GitHub account with no public email (default GitHub setting)
2. Attempt to sign in using “Sign in with GitHub”
3. Authentication fails with an error indicating that the email could not be retrieved

### Expected Behavior
If `githubUser.email` is `null`, the system should:
- Fetch the user’s email from `GET /user/emails`
- Use the primary and verified email
- Allow authentication to succeed

This aligns with GitHub OAuth best practices and common OAuth implementations.

### Actual Behavior
Authentication fails with an error similar to: 
`Unable to retrieve email from GitHub. Please ensure your email is public or grant email permissions.
`
## Additional Context
This affects users using GitHub’s default privacy settings and can block GitHub OAuth entirely in both local development and production environments.

I’m happy to submit a PR implementing this fix. Thank You.
@artahian @omegascorp @eduard-piliposyan

SargunSinghSethi:
## Summary

This PR fixes an issue where GitHub OAuth login fails for users who do not have a
public email configured on their GitHub account (default GitHub setting).

GitHub’s `/user` API can return `email: null` even when requesting the
`user:email` scope. In such cases, the user’s email must be fetched from
`/user/emails`.

---

## Problem

- GitHub accounts often have **private emails by default**
- `/user` returns `email: null`
- Current auth flow assumes `githubUser.email` is always present
- This causes GitHub OAuth login to fail for many users

---

## Solution

- If `githubUser.email` is `null`, fetch emails from `GET /user/emails`
- Select the **primary and verified** email
- Use that email to complete authentication

This follows GitHub OAuth best practices and matches behavior in common auth
solutions like NextAuth, Supabase, and Auth0.

---

## Changes

- Added support for fetching GitHub user emails via `/user/emails`
- Gracefully handles accounts with private emails
- Prevents unnecessary OAuth login failures

---

## Related Issue

Fixes #170 

---

## Notes

- No breaking changes
- Improves GitHub OAuth reliability for default GitHub accounts

@artahian @omegascorp @eduard-piliposyan 
Would be grateful if the team could look at this.
Thanks.

---

> [!NOTE]
> **Medium Risk**
> Touches the GitHub OAuth login callback and introduces an additional GitHub API dependency; failures or scope/permissions mismatches could still block login for some users.
> 
> **Overview**
> GitHub OAuth callback now resolves a usable email by falling back to `GET https://api.github.com/user/emails` when the `/user` profile has no public email, selecting the *primary verified* address.
> 
> Updates error messaging and tests to cover the extra API call and the new “no primary verified email” failure case.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b552e572e325eec8c30d71fbabe479be4c8549b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/providers/github.ts
+++ b/packages/modelence/src/auth/providers/github.ts
@@ -87,7 +87,10 @@ async function fetchGitHubUserEmails(accessToken: string): Promise<GitHubEmail[]
   return response.json();
 }
 
-async function getGitHubUserEmail(githubUser: GitHubUserInfo, accessToken: string): Promise<string | null> {
+async function getGitHubUserEmail(
+  githubUser: GitHubUserInfo,
+  accessToken: string
+): Promise<string | null> {
   if (githubUser.email) {
     return githubUser.email;
   }


packages/modelence/src/auth/providers/github.ts
```diff
@@ -99,14 +121,22 @@ async function handleGitHubAuthenticationCallback(req: Request, res: Response) {
...
-      res.status(400).json({
-        error:
-          'Unable to retrieve email from GitHub. Please ensure your email is public or grant email permissions.',
-      });
-      return;
+      const emails = await fetchGitHubUserEmails(tokenData.access_token);
+
+      const primaryVerifiedEmail = emails.find((e) => e.primary && e.verified);
+
+      githubEmail = primaryVerifiedEmail?.email ?? null;
```
omegascorp: Could you move the email selection to a separate function? I think it will make the parent function readability better

SargunSinghSethi: Sure will do that

packages/modelence/src/auth/providers/github.ts
```diff
@@ -99,14 +125,20 @@ async function handleGitHubAuthenticationCallback(req: Request, res: Response) {
...
-      });
-      return;
+      const emails = await fetchGitHubUserEmails(tokenData.access_token);
+
+      githubEmail = getPrimaryVerifiedEmail(emails);
+
+      if (!githubEmail) {
+        res.status(400).json({
+          error:
+            'Unable to retrieve a verified email from GitHub. Please ensure your GitHub account has at least one verified email.',
```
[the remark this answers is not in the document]
SargunSinghSethi: Good catch — updated the error message to explicitly mention the primary + verified email requirement so it matches the selection logic.

packages/modelence/src/auth/providers/github.ts
SargunSinghSethi: Hi @omegascorp, i have updated the code.

packages/modelence/src/auth/providers/github.ts
```diff
@@ -99,14 +125,20 @@ async function handleGitHubAuthenticationCallback(req: Request, res: Response) {
...
 
     if (!githubEmail) {
-      res.status(400).json({
-        error:
-          'Unable to retrieve email from GitHub. Please ensure your email is public or grant email permissions.',
-      });
-      return;
+      const emails = await fetchGitHubUserEmails(tokenData.access_token);
+
+      githubEmail = getPrimaryVerifiedEmail(emails);
```
omegascorp: Thank you. Could you move lines 123-133 inside the function? (The goal is to move the condition inside the function, so outside it can be just `const githubEmail = getPrimaryVerifiedEmail(something)`)

SargunSinghSethi: @omegascorp Thank you for helping me out, have updated the code.

SargunSinghSethi:
@omegascorp I’ve moved the fallback and selection logic into a helper and renamed it to getGitHubUserEmail to better reflect its responsibility. Please let me know if this looks good now. Thank you 😊

SargunSinghSethi:
@omegascorp Thank you for reviewing and testing this, I really appreciate it! 😊

PR #183, opened 2026-02-09
fix: override cloud config by local env

omegascorp:
* Override cloud configs by local env (if env variables are provided)

---

> [!NOTE]
> **Medium Risk**
> Changes config loading order for cloud-backed apps by layering local environment overrides on top of remote configs; misconfiguration could change runtime connectivity (e.g., site URL) and is exercised during periodic sync.
> 
> **Overview**
> When a remote backend is enabled, the app now loads **remote configs first, then reapplies a limited set of local env-derived overrides** (currently `MODELENCE_SITE_URL`) via a new `loadRemoteConfigs` helper used both at startup and during `startConfigSync` polling.
> 
> Extracts local env→config mapping/parsing into a new `config/local.ts` with variants for *remote vs non-remote* setups, and exposes `getSchema()` from `config/server` so sync can compute override types. Tests are updated/added to assert the new layering behavior and `loadRemoteConfigs` usage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7517ab7de62c235a1c6c0476ed81c17016a43c77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/index.test.ts
+++ b/packages/modelence/src/app/index.test.ts
@@ -526,10 +526,8 @@ describe('app/index', () => {
     expect(mockStartConfigSync).toHaveBeenCalled();
   });
 
-  test('overrides cloud config with local env when remote backend is enabled', async () => {
+  test('passes cloud configs to loadRemoteConfigs when remote backend is enabled', async () => {
     process.env.MODELENCE_SERVICE_ENDPOINT = 'https://cloud.example.com';
-    process.env.MODELENCE_SITE_URL = 'https://local.example.com';
-    process.env.MONGODB_URI = 'mongodb://localhost:27017/local';
 
     mockConnectCloudBackend.mockResolvedValue({
       configs: [
--- a/packages/modelence/src/config/sync.test.ts
+++ b/packages/modelence/src/config/sync.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, jest, test } from '@jest/globals';
+import { beforeEach, describe, expect, jest, test, afterEach } from '@jest/globals';
 
 const mockSeconds = jest.fn(() => 5000);
 const mockSyncStatus = jest.fn();
@@ -32,6 +32,7 @@ describe('config/sync', () => {
   let intervalCallback: (() => Promise<void>) | null = null;
   let intervalDelay: number | undefined;
   let startConfigSync: typeof import('./sync').startConfigSync;
+  let loadRemoteConfigs: typeof import('./sync').loadRemoteConfigs;
   const originalSetInterval = global.setInterval;
 
   beforeEach(async () => {
@@ -45,7 +46,7 @@ describe('config/sync', () => {
     }) as unknown as typeof setInterval;
     mockAcquireLock.mockResolvedValue(true as never);
     mockSyncStatus.mockResolvedValue(undefined as never);
-    ({ startConfigSync } = await import('./sync'));
+    ({ startConfigSync, loadRemoteConfigs } = await import('./sync'));
   });
 
   afterEach(() => {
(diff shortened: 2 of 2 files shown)


packages/modelence/src/app/index.ts
```diff
@@ -106,6 +107,7 @@ export async function startApp({
         stores,
       });
     loadConfigs(configs);
+    loadConfigs(getLocalConfigs(configSchema));
```
artahian: Can we just adjust how `loadConfigs` itself works? Otherwise we have to add this in more than one place and rely on the fact that it is used correctly in the right order. Didn't `loadConfigs` already load data from local configs? If so, isn't the right fix to just change the load order within `loadConfigs` itself?

But also, I'm not sure the local override is the right option, it may cause even bigger issues when you accidentally override something and keep changing in the cloud dashboard and it has no effect.

- Why was the local config not working before? If `_system.site.url` is not set, shouldn't local have been used anyway?
- For site URL specifically, can we instead add a `getSiteUrl` direct method that will fall back to `localhost:{port}` if the config is not set? The app can run on any port, but we know the port and can calculate the site URL dynamically as a fallback.

omegascorp: I think we need an option for local overrides. For example if I want to change the port to 3001 I can't do it in the cloud (it's read-only)

omegascorp: @artahian I think adding local config loading to loadConfig would be more confusing; loadConfig is really simple at the moment. The current change is pretty small and straightforward; it's repeated in sync, but the code is unlikely to change in the future.

artahian: I see, I'm just worried that this will lead to many really hard-to-debug issues in the future where you would have a local override for a config and keep changing it in the cloud and wonder why it is not working.

Instead of silently overriding locally, can we read and send the local config to the cloud backend when connecting to it? That way the cloud backend can do the override/merge itself and return the correct version (and I'd still suggest having the cloud version take priority in case of both being present). The main advantage to this would be that we can also show the local overrides in the configs dashboard and make it very clear in the UI on what's defined in the cloud and what's coming from local.

We don't necessarily have to do the UI right now, but would still be good to start sending the local configs along with the config schema when connecting.

omegascorp: I was thinking about overriding the studio configs, but in that case, the user might accidentally override the MongoDB URI and lose access to it. I'm not sure how to handle this correctly

artahian: @omegascorp ah wait, by override I didn’t mean permanently change it in the original config, I just mean separately storing the local override so Studio just knows about both its own original config and the override at the same time, so it can explicitly show in the dashboard / UI what's going on and what's the current effective value.

The UI part is more of an improvement for the future, for now my only suggestion is to send local config overrides to Studio and let it do the merging, so it can also know what's the final value and display that in the UI later. Otherwise in the future when you open your Studio app configs page you will not see anything about local overrides and will think your app is still using the old value even when it has a local override.

omegascorp: @artahian ah, now I got it

packages/modelence/src/config/sync.ts
```diff
@@ -34,4 +35,5 @@ export function startConfigSync() {
 async function syncConfig() {
   const { configs } = await fetchConfigs();
   loadConfigs(configs);
+  loadConfigs(getLocalConfigs(getSchema(), 'withRemoteServer'));
```
artahian, line 38: Minor: I'd recommend merging these two commands into a `loadRemoteConfigs` or similar function to make sure it's always doing this exact combination wherever it is used, in case there are more usages later.

PR #184, opened 2026-02-11
feat(auth) expose name and picture in useSession() user

AtulYadav25:
This PR fixes the authentication flow by exposing `name` and `picture` fields on the `user` object returned from:

```ts
import { useSession } from 'modelence/client';
```

Previously, although OAuth providers (Google/GitHub) returned profile data such as name and avatar, these fields were not:

* Stored in the database
* Included in the `UserInfo` TypeScript type

As a result, accessing `user.name` or `user.picture` caused TypeScript errors.

This PR resolves that issue.

### 1. OAuth Providers (Google & GitHub)

* Fetch `name` and avatar (`picture` / `avatar_url`) from provider APIs
* Pass them into the authentication pipeline

I tried doing console log of users data fetched from the providers
Result
Screenshot (519)
Screenshot (520)

### 2. Database Schema Update

* Added `name` and `picture` fields to user schema
* On re-login, user profile data is updated if changed

### 3. Authentication Flow Update

* `authenticate` now returns `name` and `picture` also
* `getOwnProfile` includes profile fields of `name` and `picture`

### 4. Type Updates

Updated `UserInfo` type:

```ts
export type UserInfo = {
  id: string;
  handle: string;
  roles: string[];
  hasRole: (role: string) => boolean;
  requireRole: (role: string) => void;
  name?: string;
  picture?: string;
};
```

### 5. Client Session Update

* Updated `packages/modelence/src/client/session.ts`
* Extended Zod schema to validate `name` and `picture` optionally
* `useSession()` now properly exposes them

### 6. auth-ui Update

* Updated `UserProfile` component
* Profile page now displays user avatar and name

---

## Testing

* Built package successfully using `pnpm build`
* Packed locally using `pnpm pack`
* Installed in test project
* Verified:
  * Google login returns `name` and `picture`
  * GitHub login returns `name` and `avatar_url`
  * Data stored in MongoDB
  * `useSession()` includes required fields
  * Avatar and name render correctly in UI

Screenshot are attached: useSession() use returns the `name` and `picture` :
Screenshot (527)

---

> [!NOTE]
> **Medium Risk**
> Touches core authentication/signup/OAuth linking and introduces new DB fields and handle resolution logic, which could affect account creation/login behavior and uniqueness constraints if edge cases aren’t covered.
> 
> **Overview**
> Modelence auth now persists and exposes user profile fields (`firstName`, `lastName`, `avatarUrl`) end-to-end: stored in `_modelenceUsers`, returned from `authenticate`/login, included in `getOwnProfile`, and surfaced in `useSession()`/client `UserInfo`.
> 
> OAuth (Google/GitHub) now extracts these fields from provider responses and **backfills missing profile fields** for existing users on login/auto-link, while new users get a unique handle via a new `AuthConfig.generateHandle` hook (with automatic conflict suffixing) or email-derived fallback.
> 
> Email/password signup is extended to accept optional profile fields + custom `handle`, runs a new `AuthConfig.validateSignup` hook, resolves unique handles via new `auth/utils.ts`, and a new `updateProfile` mutation/client API adds validated profile edits with optional `validateProfileUpdate`. Docs and `@modelence/auth-ui` `UserProfile` are updated to reflect/show these profile fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 80c4ebda114677c5356cb226cd5108f5ff3c0437. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/utils.ts
+++ b/packages/modelence/src/auth/utils.ts
@@ -1,3 +1,4 @@
+import { randomBytes } from 'crypto';
 import { usersCollection } from './db';
 import { validateHandle, MAX_HANDLE_LENGTH, MIN_HANDLE_LENGTH } from './validators';
 
@@ -10,31 +11,66 @@ async function findAvailableHandle(baseHandle: string): Promise<string> {
   // Truncate base handle to MAX_HANDLE_LENGTH so the unsuffixed form is valid.
   const truncatedBase = baseHandle.slice(0, MAX_HANDLE_LENGTH);
 
-  const firstCheck = await usersCollection.findOne(
-    { handle: truncatedBase },
-    { collation: { locale: 'en', strength: 2 } }
-  );
-
-  if (!firstCheck) {
-    return truncatedBase;
-  }
-
-  let suffix = 2;
-  while (true) {
-    const suffixStr = `_${suffix}`;
-    // Truncate base so that base + suffix fits within the limit.
-    const candidate = `${truncatedBase.slice(0, MAX_HANDLE_LENGTH - suffixStr.length)}${suffixStr}`;
-    const conflict = await usersCollection.findOne(
-      { handle: candidate },
+  // Check the unsuffixed base handle first.
+  try {
+    const firstCheck = await usersCollection.findOne(
+      { handle: truncatedBase },
       { collation: { locale: 'en', strength: 2 } }
     );
 
-    if (!conflict) {
-      return candidate;
+    if (!firstCheck) {
+      return truncatedBase;
     }
-
-    suffix++;
+  } catch (err) {
+    throw new Error(`Database error while checking handle availability: ${err}`);
   }
+
+  // Try sequential suffixes _2 through _51 (50 attempts).
+  const MAX_SUFFIX_VALUE = 51;
+
+  for (let suffix = 2; suffix <= MAX_SUFFIX_VALUE; suffix++) {
+    const suffixStr = `_${suffix}`;
+    const candidate = `${truncatedBase.slice(0, MAX_HANDLE_LENGTH - suffixStr.length)}${suffixStr}`;
+
+    try {
+      const conflict = await usersCollection.findOne(
+        { handle: candidate },
+        { collation: { locale: 'en', strength: 2 } }
+      );
+
+      if (!conflict) {
+        return candidate;
+      }
+    } catch (err) {
+      throw new Error(`Database error while checking handle "${candidate}": ${err}`);
+    }
+  }
+
+  // Fallback: sequential suffixes exhausted — use random hex suffixes.
+  // Limit the number of attempts to avoid an infinite loop in case of persistent DB issues
+  const MAX_RANDOM_ATTEMPTS = 10;
+
+  for (let attempt = 0; attempt < MAX_RANDOM_ATTEMPTS; attempt++) {
+    const randomSuffix = `_${randomBytes(3).toString('hex')}`;
+    const candidate = `${truncatedBase.slice(0, MAX_HANDLE_LENGTH - randomSuffix.length)}${randomSuffix}`;
+
+    try {
+      const conflict = await usersCollection.findOne(
+        { handle: candidate },
+        { collation: { locale: 'en', strength: 2 } }
+      );
+
+      if (!conflict) {
+        return candidate;
+      }
+    } catch (err) {
+      throw new Error(`Database error while checking handle "${candidate}": ${err}`);
+    }
+  }
+
+  throw new Error(
+    `Could not generate a unique handle for base "${baseHandle}" after exhausting all attempts.`
+  );
 }
 
 /**


packages/auth-ui/src/components/UserProfile.tsx
```diff
@@ -168,13 +168,15 @@ type ProfileData = {
   handle: string;
   emails: { address: string; verified: boolean }[];
   authMethods: string[];
+  name?: string;
+  picture?: string;
```
omegascorp: avatarUrl - is more common name for this

packages/auth-ui/src/components/UserProfile.tsx
```diff
@@ -168,13 +168,15 @@ type ProfileData = {
   handle: string;
   emails: { address: string; verified: boolean }[];
   authMethods: string[];
+  name?: string;
```
omegascorp: I would suggest having `firstName` and `lastName` instead.

omegascorp: For github we can just split for now (it might be wrong in some cases, but I don't see a better solution)

omegascorp (commented):
Thank you for the PR, I've added some comments, there are also some form the bot

artahian:
> @omegascorp @artahian
> Do we need the `handle` field even when `email` already exists on the user object?
> 
> Alternatively, I think `handle` should represent something more user focus like:
> 
> GitHub username for GitHub OAuth Full name from Google OAuth
> 
> I’d like to clarify the real role of `handle` before making next changes.

Good question - and we've intentionally picked `handle` as the name for this field and not `username` so it can be more generic. Some apps/platforms (like Modelence Cloud itself) don't really need a "username" concept and it only works with emails. I think a username is mostly used for social apps, where you can see and interact with other users - e.g. Discord and even Spotify.

Ideally I'd like `handle` to remain flexible - so I'd suggest to add support for a custom callback where the user can define how to generate the handle for each user signup (and even better if it is smart enough to resolve conflicts on its own and automatically adding `_2`, _3`, etc as suffixes on collisions with existing handles.

> 
> I reviewed the same for `googleId` and `githubId` field, they both aren't sent in user object as mentioned in documentation.

@omegascorp I'm not sure about this part, deferring to you. But @AtulYadav25 what's the use case for having `googleId` and `githubId` sent to the frontend automatically? `useSession` is basically made for having your current user's basic fields readily available at all times. For anything more specific you can always have a custom method that fetches these on demand.

omegascorp:
@AtulYadav25, the documentation is wrong there; the GitHub ID is stored in `authMethods.github.id`, we need to update the docs to reflect the real user object that is published

AtulYadav25:
@omegascorp
To make the authentication workflow cleaner and more consistent (and considering @artahian suggestions), I’ve few improvements.

While these are the current updates
- I added `firstName`, `lastName`, and `avatarUrl` to the user schema.
- For OAuth users, these fields are automatically populated from the provider.
- For non-OAuth users (password users), there is currently no good way to set `firstName` and `lastName` or `avatarUrl`.

This can be an inconsistency in user data
OAuth users may have profile information.
Password users do not have an official way to set it.

## Possible Improvements
1. Add a `generateHandle` callback
This would allow developers to customize how handles are generated during signup.

However, even with this, developers still wouldn’t have a clean way to:
Update `handle`
Set or update `firstName`, `lastName`, or `avatarUrl` later

That makes the framework feel a bit rigid.

2. Introduce a `handleUpdateProfile` function in _system.user module mutation

This function would allow updating:
`handle`, `firstName`, `lastName`, `avatarUrl`

Also, we could update `handleSignUpWithPassword` to accept these `handle`, `firstName`, `lastName`, `avatarUrl` fields as OPTIONAL during user creation.

This:
Keep OAuth and password users consistent in terms of profile data
Allow profile updates in a clear and structured way

3. Add `validateSignUp` callback in `authConfig`

Introduce a `validateSignUp` callback that runs during user creation. This allows developers to enforce custom validation rules without modifying the core authentication flow.

For example:

```ts
auth: {
  validateSignUp: ({ firstName, lastName, handle }) => {
    if (!firstName) {
      throw new Error("FirstName is required");
    }
  }
}
```

This callback would be executed inside `handleSignUpWithPassword`.

---

### NEXT?

To move forward, I have two approaches:

Add only the `generateHandle` callback for now and keep the current structure otherwise unchanged.

OR
Introduce a proper `handleUpdateProfile` function and extend `handleSignUpWithPassword` to accept optional profile fields (`firstName`, `lastName`, `avatarUrl`, `handle`), along with adding the `validateSignUp` callback.


I’m open to suggestions if there’s a better way to do this or more improvements.

packages/modelence/src/app/authConfig.ts
```diff
@@ -1,4 +1,19 @@
 import { AuthErrorProps, AuthSuccessProps, User } from '@/auth/types';
+import { UpdateProfileArgs } from '@/methods/types';
+
+type GenerateHandleInput = {
+  email: string;
+  firstName?: string;
+  lastName?: string;
+};
+
+type signUpArgs = {
```
omegascorp: Should be `SignUpProps` to match other types

packages/modelence/src/app/authConfig.ts
```diff
@@ -1,4 +1,19 @@
 import { AuthErrorProps, AuthSuccessProps, User } from '@/auth/types';
+import { UpdateProfileArgs } from '@/methods/types';
+
+type GenerateHandleInput = {
```
omegascorp: Should be `GenerateHandleProps`

packages/modelence/src/app/authConfig.ts
```diff
@@ -1,4 +1,19 @@
...
+  lastName?: string;
+};
+
+type signUpArgs = {
+  email: string;
+  firstName?: string;
+  lastName?: string;
+  password: string;
+  handle?: string;
+};
```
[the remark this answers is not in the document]
omegascorp: This is a valid comment

packages/modelence/src/app/authConfig.ts
```diff
@@ -33,19 +54,33 @@ export type AuthOption = {
...
+ *       // Generate handle
+ *       return 'user123';
+ *     },
  *   }
  * });
  * ```
  */
 export type AuthConfig = {
+  // Optional pre-signup validation hook.
+  validateSignUp?: (props: signUpArgs) => void | Promise<void>;
```
omegascorp: Signup as a noun in other places with lowercase u, let's keep it consistent

packages/modelence/src/auth/providers/oauth-common.ts
```diff
@@ -5,12 +5,16 @@ import { createSession } from '@/auth/session';
...
+import { resolveUniqueHandle } from '../utils';
 
 export interface OAuthUserData {
   id: string;
   email: string;
   emailVerified: boolean;
   providerName: 'google' | 'github';
+  firstName?: string;
+  lastName?: string;
+  avatarUrl?: string;
```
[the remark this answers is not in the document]
omegascorp: I don't think we need to sync existing data since the user might have multiple OAuth providers connected, and the user can also update data manually. But if there is no firstName, lastName, and avatarUrl in DB, we can set it

packages/modelence/src/auth/profile.ts
```diff
@@ -12,5 +14,56 @@ export async function getOwnProfile(_args: Args, { user }: Context) {
     handle: profile.handle,
     emails: profile.emails,
     authMethods: Object.keys(profile.authMethods || {}),
+    firstName: profile.firstName ?? undefined,
+    lastName: profile.lastName ?? undefined,
+    avatarUrl: profile.avatarUrl ?? undefined,
+  };
+}
+
+export async function handleUpdateProfile(_args: UpdateProfileArgs, { user }: Context) {
```
omegascorp: There should be no underscore before args here, also args should be renamed to props for consistancy

packages/modelence/src/auth/signup.ts
```diff
@@ -5,13 +5,15 @@ import { usersCollection } from './db';
 import { isDisposableEmail } from './disposableEmails';
 import { consumeRateLimit } from '../rate-limit/rules';
 import { sendVerificationEmail } from './verification';
-import { validateEmail, validatePassword } from './validators';
+import { validateEmail, validatePassword, validateProfileFields } from './validators';
 import { getAuthConfig } from '@/app/authConfig';
+import { resolveUniqueHandle } from './utils';
 
 export async function handleSignupWithPassword(
   args: Args,
```
omegascorp: Let's rename this to props and set a proper type here

packages/modelence/src/methods/types.ts
```diff
@@ -27,6 +27,13 @@ export type Context = {
 
 export type Args = Record<string, unknown>;
 
+export type UpdateProfileArgs = {
```
omegascorp: Let's rename to UpdateProfileProps

packages/modelence/src/auth/profile.ts
AtulYadav25: Should I replace all occurrences of  `Args` with `Props` ?

Here , `export type Args = Record<string, unknown>;` in file `modelence/src/methods/types.ts`

omegascorp: I would suggest replacing only the parts that are changed. `Args` is used in many places.

packages/modelence/src/auth/validators.ts
```diff
@@ -7,3 +8,68 @@ export function validatePassword(value: string) {
...
+ * Validates and trims profile fields against the defined rules.
+ *
+ * For each field present in `fields`:
+ * - Trims whitespace
+ * - Sets fields to `undefined` when cleared (empty string)
+ * - Enforces min/max length constraints
+ *
+ * @returns An object containing only the validated, trimmed fields that were provided.
+ */
+export function validateProfileFields(
```
omegascorp, line 43: Can we replace this via `zod` validation?

omegascorp:
@AtulYadav25 thank you, I've added one more comment

packages/modelence/src/auth/signup.ts
```diff
@@ -58,11 +63,43 @@ export async function handleSignupWithPassword(
...
+    // Resolve unique handle
+    let resolvedHandle: string;
+
+    if (profileFields.handle) {
+      resolvedHandle = await resolveUniqueHandle(profileFields.handle, email);
+    } else if (authConfig.generateHandle) {
+      const generated = await authConfig.generateHandle({
+        email,
+        ...profileFields,
+      });
```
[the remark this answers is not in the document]
omegascorp: @AtulYadav25, could you check the cursor comments? This one is relevant.

packages/modelence/src/auth/validators.ts
omegascorp: @AtulYadav25, could you check this comment?

packages/modelence/src/auth/signup.ts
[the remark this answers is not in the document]
AtulYadav25: > @AtulYadav25, could you check the cursor comments? This one is relevant.

Yup, I did a fix on this, maybe it is not visible because you might be seeing some limited last commits.
Could you please check through this `all commits`
Screenshot (604)


For a quick review
What i made changes:

From this:
```ts
 const generated = await authConfig.generateHandle({
        email,
        ...profileFields,
      });
```
To this
```ts
const generated = await authConfig.generateHandle({
        email,
        firstName: profileFields.firstName,
        lastName: profileFields.lastName,
      });
```

packages/modelence/src/auth/validators.ts
AtulYadav25: > @AtulYadav25, could you check this comment?

yes, I used `zod` and made this more clean in the latest commit 😅

omegascorp (approved):
Thank you @AtulYadav25, appoved.

PR #191, opened 2026-02-18
feat(auth): add ip & email rate limiting to password reset endpoint

SargunSinghSethi:
# Add IP & Email-Based Rate Limiting to Password Reset

## Overview

This PR enhances the password reset flow by introducing **email-based rate limiting in addition to existing IP-based protection**.

Previously, password reset requests were either unprotected or only IP-protected. This allowed attackers to bypass limits by rotating IP addresses and repeatedly triggering reset emails for a target address.

This update provides layered protection at both the requester and target levels.

------------------------------------------------------------------------

### 1️⃣ Extended Rate Limit Types

Added `'email'` to the `RateLimitType` union:

``` ts
export type RateLimitType = 'ip' | 'user' | 'email';
```

Updated the database schema to allow `'email'` as a valid rate limit type.

------------------------------------------------------------------------

### 2️⃣ Added Email-Based Rate Limit Rules

Registered new rules under the `password-reset-email` bucket:

-   **5 requests per hour per email**
-   **10 requests per day per email**

``` ts
{
  bucket: 'password-reset-email',
  type: 'email',
  window: time.hours(1),
  limit: 5,
},
{
  bucket: 'password-reset-email',
  type: 'email',
  window: time.days(1),
  limit: 10,
},
```

------------------------------------------------------------------------

### 3️⃣ Enforced Email Rate Limiting in Reset Flow

Inside `handleSendResetPasswordToken`:

``` ts
const email = validateEmail(args.email as string);

await consumeRateLimit({
  bucket: 'password-reset-email',
  type: 'email',
  value: email,
});
```

This ensures:

-   Rate limiting applies even if attackers rotate IP addresses
-   Abuse is controlled at the target email level
-   The system fails fast before sending any email

------------------------------------------------------------------------

## Security Improvements

Without this change, attackers could:

-   Perform email bombing attacks
-   Spam password reset emails
-   Abuse email infrastructure
-   Degrade user experience

This PR significantly improves the authentication system's security posture.

------------------------------------------------------------------------

## Tests

-   Added test coverage to verify `consumeRateLimit()` is called with
    `{ type: 'email', value: email }`
-   Updated mocks to support new `time.hours()` helper
-   Ensured rate limit rules are properly registered

------------------------------------------------------------------------

## Impact

-   Prevents password reset abuse
-   Protects users from targeted email spam
-   Uses existing rate limiting infrastructure
-   No breaking changes

------------------------------------------------------------------------

## Result

Password reset flow is now protected by:

-   IP-based rate limiting (requester-level protection)
-   Email-based rate limiting (target-level protection)

This provides layered, production-grade abuse prevention.

Would be grateful if the team could look at this.
Thank you 😊
@artahian  @omegascorp  @eduard-piliposyan 

---

> [!NOTE]
> **Medium Risk**
> Touches password-reset and rate-limit persistence paths; incorrect limits or key normalization could block legitimate resets or reduce abuse protection.
> 
> **Overview**
> Adds layered rate limiting to `handleSendResetPasswordToken` by consuming `passwordReset` limits per-IP (when available) and per-email before generating/storing a token or sending mail.
> 
> Extends the rate-limit system to support an `email` identifier type (types + DB schema) and registers new `passwordReset` rules in `auth/user.ts` (IP: 10/15m, 100/day; email: 5/hour, 10/day). Also normalizes emails to lowercase in `validateEmail` and updates email lookups in login/signup/verification accordingly, with tests updated to cover the new rate-limit calls and `time.hours()` usage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 433fbb5ec7fedf19a40b564118feb50ab5225baa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/login.ts
+++ b/packages/modelence/src/auth/login.ts
@@ -48,7 +48,7 @@ export async function handleLoginWithPassword(
       throw incorrectCredentialsError();
     }
 
-    const emailDoc = userDoc.emails?.find((e) => e.address === email);
+    const emailDoc = userDoc.emails?.find((e) => e.address.toLowerCase() === email);
 
     if (!emailDoc?.verified && getEmailConfig()?.provider) {
       if (ip) {
--- a/packages/modelence/src/auth/signup.ts
+++ b/packages/modelence/src/auth/signup.ts
@@ -41,7 +41,7 @@ export async function handleSignupWithPassword(
     );
 
     if (existingUser) {
-      const existingEmail = existingUser.emails?.find((e) => e.address === email);
+      const existingEmail = existingUser.emails?.find((e) => e.address.toLowerCase() === email);
       if (existingUser.status === 'disabled') {
         throw new Error(
           `User is marked for deletion, please contact support if you want to restore the account.`
--- a/packages/modelence/src/auth/verification.ts
+++ b/packages/modelence/src/auth/verification.ts
@@ -177,7 +177,7 @@ export async function handleResendEmailVerification(args: Args, { connectionInfo
     return resendVerificationResponse;
   }
 
-  const emailDoc = userDoc.emails?.find((e) => e.address === email);
+  const emailDoc = userDoc.emails?.find((e) => e.address.toLowerCase() === email);
 
   if (!emailDoc || emailDoc.verified) {
     return resendVerificationResponse;


hamazaspintegral:
Hi, everting looks ok. Thanks for asking.

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -39,8 +40,23 @@ const passwordResetSent = {
...
+    await consumeRateLimit({
+      bucket: 'password-reset',
+      type: 'ip',
+      value: ip,
+    });
+  }
   const email = validateEmail(args.email as string);
 
+  await consumeRateLimit({
+    bucket: 'password-reset-email',
```
artahian: Why use a different bucket name here and not just `password-reset`?

packages/modelence/src/auth/user.ts
```diff
@@ -105,6 +105,30 @@ export default new Module('_system.user', {
       window: time.days(1),
       limit: 10,
     },
+    {
+      bucket: 'password-reset',
```
artahian: In other places we use camelCase, can we change this to `passwordReset`?

SargunSinghSethi: Than you for the feedback @artahian 
I will get on with the changes. Thank you 😊.

SargunSinghSethi:
@artahian @omegascorp 
I’ve updated the implementation to use a single passwordReset bucket (camelCase for consistency with other buckets).

Both IP-based and email-based limits now use the same bucket and are differentiated using type: 'ip' and type: 'email'.

I’ve also updated the related tests accordingly.

Thank you 😊.

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -39,8 +40,23 @@ const passwordResetSent = {
...
 export async function handleSendResetPasswordToken(args: Args, { connectionInfo }: Context) {
+  const ip = connectionInfo?.ip;
+
+  if (ip) {
+    await consumeRateLimit({
+      bucket: 'passwordReset',
+      type: 'ip',
+      value: ip,
+    });
+  }
```
artahian, line 52: I think we need to move both consume calls below, to happen after other validation checks. Otherwise we're counting all attempts, including the ones where there was an error and no reset password has been attempted. If a user tries to reset and fails validation multiple times there is a chance they will be locked out completely from even sending one real reset password email.

SargunSinghSethi: That makes sense @artahian 👍
I’ll move both consumeRateLimit calls to occur after email validation, so invalid input doesn’t consume rate limits, while still protecting actual reset attempts.
Thanks for pointing that out 🙂

omegascorp (approved):
Thank you, @SargunSinghSethi, approved

SargunSinghSethi:
@artahian @omegascorp

I’ve moved both consumeRateLimit calls to occur after email validation, so invalid inputs no longer consume rate limits while valid reset attempts remain protected.

I also addressed the email case-sensitivity issue more comprehensively:

- Emails are now normalized to lowercase in validateEmail to prevent rate limit bypass via case variations.
- Updated in-memory comparisons in login.ts, signup.ts, and verification.ts to use lowercase matching, aligning with the database’s case-insensitive collation.

This prevents both the rate-limit bypass and login regressions for users with mixed-case stored emails.

All related tests are passing.

Thank you 😊

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -40,6 +41,21 @@ const passwordResetSent = {
...
+
+  await consumeRateLimit({
+    bucket: 'passwordReset',
+    type: 'email',
+    value: email,
+  });
```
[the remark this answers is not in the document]
SargunSinghSethi: Thanks for the review.

In this case, the current placement of the rate limits is intentional to preserve consistent behavior regardless of whether the email exists in the system.

If rate limiting were applied only after confirming a user exists, it would introduce a user enumeration vector (existing emails would eventually return 429, while non-existing ones would not).

Applying rate limits uniformly to all valid reset attempts maintains the anti-enumeration property while still protecting against abuse.

@artahian @omegascorp — does this align with the intended security model, or would you prefer a different tradeoff here? Happy to adjust based on your guidance.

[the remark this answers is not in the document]
omegascorp: The current approach is safer, let's procced with it

PR #195, opened 2026-02-20
fix: create indexes before running startApp

omegascorp:
* Ensure `blocking` indexes are fully created before cron jobs
* Warn and continue starting app when index creation is falied

---

> [!NOTE]
> **Medium Risk**
> Changes the startup sequencing around DB index creation/migrations/cron, which can affect boot timing and runtime behavior; failures now warn-and-continue, so missed indexes could impact lock/migration correctness if not monitored.
> 
> **Overview**
> Adds a per-store `indexCreationMode` (default `background`) to control whether `Store.createIndexes()` runs asynchronously during startup or *blocks* startup until completion.
> 
> Updates `startApp` to await all `blocking` stores’ index creation before starting migrations/cron, then kicks off `background` index creation, logging a warning (but continuing startup) if index creation fails. Applies `blocking` mode to the internal `_modelenceLocks` store and updates docs/tests accordingly; bumps `modelence` to `0.12.3`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3c529dc86d3c40845c91aac23840d39c001cad3f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/index.ts
+++ b/packages/modelence/src/app/index.ts
@@ -119,11 +119,7 @@ export async function startApp({
   if (mongodbUri) {
     await connect();
     initStores(stores);
-  }
 
-  startMigrations(migrations);
-
-  if (mongodbUri) {
     for (const store of stores) {
       if (store.getIndexCreationMode() === 'blocking') {
         await createStoreIndexes(store);
@@ -136,6 +132,8 @@ export async function startApp({
       }
     }
   }
+
+  startMigrations(migrations);
 
   if (hasRemoteBackend) {
     await initMetrics();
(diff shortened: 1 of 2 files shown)


packages/modelence/src/app/index.ts
```diff
@@ -119,16 +119,11 @@ export async function startApp({
   if (mongodbUri) {
     await connect();
     initStores(stores);
+    await Promise.all(stores.map((store) => store.createIndexes()));
```
artahian: @omegascorp I think the issue is different though, it's not just about migrations.

- For one, this was working fine before the `_modelence` index prefix management. Even with the old order.
- Also, the issue is with crons as well, not just migration locks. I assume the `await` you added may fix this, but isn't this going to take a long time to create?

PR #236, opened 2026-03-30
fix(store): extended store merging

omegascorp:
* merge indexes of the extended stores before creating
* merge fields of extended stores before sending to the server

---

> [!NOTE]
> **Medium Risk**
> Changes how `Store.extend()` chains are tracked and how app startup/cloud-connect choose which store definitions drive index reconciliation and remote metadata. Mistakes could lead to missing/incorrect indexes or schema/index metadata being sent to Modelence Cloud.
> 
> **Overview**
> Fixes extended-store merging by introducing a linear *extension chain* for `Store.extend()` and selecting the chain tail as the single source of truth.
> 
> App startup now calls new `resolveStores()` to dedupe stores, detect same-collection collisions, initialize all unique store instances, but run index reconciliation/migrations and `connectCloudBackend` metadata sync using only the **effective (tail) stores**.
> 
> `connectCloudBackend` now accepts optional stores and sends richer model metadata (serialized schema plus `indexes`, `searchIndexes`, and `indexCreationMode`). Tests were updated/added to cover chain behavior, collision detection, and the revised startup/cloud payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a230634d0a0a38beeba9dbc2ddd0c092a9349353. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


artahian (commented):
@omegascorp I'm a little unsure about the merging part, is there a simpler way to do this? For example, why can't we merge on Store.extend, which explicitly says B is overwriting A, instead of then searching through the whole group and merging from scratch?

Also, can we just ignore the original instead of merging? If we defined an extended users collection, I think that should invalidate the original one since it's already using that and adding more things on top.

And in general do we allow extending the same store twice? I think you should only be able to extend Store A once, so it would fail if you call Store.extend twice on the same exact Store (we can still allow to chain it, but not branch out with more than one branch from the same).

And if we do this, then can't we just set "sourceStore" or something like that in the extended store, and later in `startApp` simply filter out all stores which are set as "sourceStore" on another store?

packages/modelence/src/data/store.ts
```diff
@@ -429,6 +447,39 @@ export class Store<
...
+    while (current._chainParent) {
+      current = current._chainParent;
+    }
+    return current;
+  }
+
+  /** @internal */
+  isInSameChain(other: AnyStore): boolean {
+    return this.getChainRoot() === other.getChainRoot();
+  }
```
artahian: Do we really need this and the `getChainRoot` method? I don't see where it's being used (hopefully we don't need it given that we are going with a simple approach).

packages/modelence/src/app/backendApi.ts
```diff
@@ -44,14 +49,25 @@ export async function connectCloudBackend({
...
+            indexes: merged.indexes,
+            searchIndexes: merged.searchIndexes,
+            indexCreationMode: merged.indexCreationMode,
+          }))
+        : Object.values(stores).map((store) => ({
+            name: store.getName(),
+            schema: store.getSerializedSchema(),
+            collections: [store.getName()],
+            version: 2,
+          }));
```
artahian: Can we just call them stores instead of models? Also, why do we need the branching here - can't we always take effective stores? As far as I understand we can simply ignore all stores that have been extended, no?

Basically:
- Filter out all stores that are not the final extended version (i.e. have been extended by another store). This leaves only the most complete Store for each collection.
- When provisioning, collect indexes not just from current object but from all source stores recursively.

packages/modelence/src/data/store.ts
omegascorp: Removed isInSameChain since it was a leftover, but we still need getChainRoot

packages/modelence/src/app/backendApi.ts
omegascorp: Branching was a leftover from the old implementation, updated

omegascorp:
@artahian please check again

artahian (approved):

PR #245, opened 2026-04-25
feat(auth): make authentication rate limits configurable via AuthConfig

### LINKED ISSUE #31: Make signup rate limits configurable
Currently, our rate limits for user signups are hardcoded:
- 20 signups within 15 mins from same IP
- 200 signups within 24 hours from same IP

Ideally, this should be configurable and use Modelence config, which will later be controlled by dedicated user/auth settings UI in Modelence Cloud.

Abhijeet1005:
Closes #31

Previously, all auth rate limits (signup, signupAttempt, signin, verification, passwordReset) were hardcoded as static values inside the _system.user Module constructor, making them impossible to tune without forking the framework.

This change introduces AuthRateLimitsConfig — a typed, fully-documented config type that lets users override any per-action limit through the existing startApp auth config:

  startApp({
    auth: {
      rateLimits: {
        signup: { perIp15Minutes: 5, perIpPerDay: 50 },
        signin: { perIp15Minutes: 10 },
      },
    },
  });

Unspecified fields fall back to the original hardcoded defaults, so there is no behaviour change for existing apps.

Implementation notes:
- Auth rate limits are moved out of the Module constructor into an exported buildAuthRateLimits(config?) builder in auth/user.ts. This sidesteps the init-order issue: initRateLimits is called in startApp before setAuthConfig, so reading from getAuthConfig() at that point would return an empty object. Passing auth.rateLimits directly from the startApp parameter avoids the race entirely.
- AuthRateLimitsConfig is re-exported from modelence/server so consumers can type their config without reaching into internal paths.
- 5 new unit tests cover defaults, full overrides, partial overrides, multi-bucket overrides, and rule count invariant.
- fix: add shebang to .husky/pre-commit so Git can spawn it on Windows.

---

> [!NOTE]
> **Medium Risk**
> Touches authentication rate-limiting behavior and startup wiring; misconfiguration or merge bugs could weaken brute-force protection or change effective limits at runtime, though defaults remain unchanged when unset.
> 
> **Overview**
> Adds `AuthRateLimitsConfig` to `AuthConfig` so apps can override per-endpoint auth rate limits (signup/signin/verification/password reset) while preserving defaults for unspecified fields.
> 
> Refactors auth rate limits into an exported `buildAuthRateLimits()` in `auth/user.ts` and updates `startApp` to apply overrides by mutating `userModule.rateLimits` before rate-limit collection/initialization; `Module.rateLimits` is made mutable to support this. Updates tests to cover default/override behavior and ensures the effective rules are surfaced via the `_system.user` module, and re-exports `AuthRateLimitsConfig` from `server.ts`. Also fixes the `.husky/pre-commit` hook by adding a shebang.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6eda501590426092d33495a974376eef699c0ee0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/user.ts
+++ b/packages/modelence/src/auth/user.ts
@@ -1,6 +1,8 @@
 import { randomBytes } from 'crypto';
 
+import type { AuthRateLimitsConfig } from '../app/authConfig';
 import { Module } from '../app/module';
+import { RateLimitRule } from '../rate-limit/types';
 import { time } from '../time';
 import {
   dbDisposableEmailDomains,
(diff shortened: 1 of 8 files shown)


omegascorp (commented):
@Abhijeet1005, this looks good overall, but `buildAuthRateLimits` should still be passed to the user module since we might read it later from modules and display it in the Modelence Cloud

packages/modelence/src/auth/user.ts
```diff
@@ -16,123 +18,137 @@ import { handleSignupWithPassword } from './signup';
...
+    resetPassword: handleResetPassword,
+    updateProfile: handleUpdateProfile,
+    unlinkOAuthProvider: handleUnlinkOAuthProvider,
+  },
+  cronJobs: {
+    updateDisposableEmailList: updateDisposableEmailListCron,
+  },
+  // Default auth rate limits — replaced at startApp time with any user
+  // overrides from `auth.rateLimits`. Keeping the rules on the module makes
+  // them discoverable for module-level introspection (e.g. Modelence Cloud).
```
omegascorp: Please remove this comment (all 3 lines)

Abhijeet1005:
> @Abhijeet1005, this looks good overall, but `buildAuthRateLimits` should still be passed to the user module since we might read it later from modules and display it in the Modelence Cloud

Pushed some improvements, please share your thoughts

packages/modelence/src/auth/user.ts
Abhijeet1005: done

omegascorp (approved):
@Abhijeet1005 thank you, looks good

PR #246, opened 2026-04-27
fix: improve sanitization performance

omegascorp:
Scenario | OLD | NEW | Speedup
-- | -- | -- | --
Primitive array (1 000 numbers, no ObjectId) | 161K ops/s | 2.75M ops/s | 17×
Array of 500 plain objects, no ObjectId | 4.5K ops/s | 11.4K ops/s | 2.5×
Array of 500 docs with ObjectId _id (real work) | 4.5K ops/s | 7.8K ops/s | 1.7×
Deeply nested object, depth 10, no ObjectId | 368K ops/s | 778K ops/s | 2.1×
Realistic mixed doc (some ObjectIds) | 587K ops/s | 1.05M ops/s | 1.8×

---

> [!NOTE]
> **Medium Risk**
> Changes core response sanitization used in method and live-query responses; while behavior should be equivalent, subtle edge cases in traversal/copying could affect returned data shapes or mutation expectations.
> 
> **Overview**
> Improves `sanitizeResult` performance by switching from always-deep-mapping to a **copy-on-write** traversal that returns the original array/object reference when no `ObjectId` conversions are needed.
> 
> Adds a small benchmark harness (`npm run bench`) plus a `serialize.bench.ts` suite to measure sanitization vs `JSON.stringify`, and extends tests to assert reference-preserving behavior and minimal copying when only some fields contain `ObjectId`s. Bumps `modelence` version to `0.15.4`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 273526970eb9d3e376f5cda0a6d1a64abe760801. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.15.3",
+  "version": "0.15.4",
   "description": "The Node.js Framework for Real-Time MongoDB Apps",
   "main": "dist/index.js",
   "types": "dist/global.d.ts",
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -31,6 +31,7 @@
     "lint": "eslint src --ext .ts,.tsx --fix",
     "lint:check": "eslint src --ext .ts,.tsx",
     "prepublishOnly": "npm run build",
+    "bench": "tsx src/bench/index.ts",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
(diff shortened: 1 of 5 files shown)


artahian (commented):
@omegascorp the difference in performance on real data seems small, do you have the actual numbers of how long the transformation takes on a real app builder chat data? We will most likely have the same issue on production if the difference is just 2x.

omegascorp:
I think it's the reason why the production was slow. I made a change in the studio to avoid large data transfers

eduard-piliposyan (approved):

PR #248, opened 2026-05-04
feat: ssr support

omegascorp:
* SSR support
* `modelenceQuery` is moved to the `modelence` package
* `@tanstack/react-query` is converted to a peer dependency for `modelence`

---

> [!NOTE]
> **High Risk**
> Large SSR surface area plus auth cookie and session reconciliation logic; misconfiguration can cause hydration mismatches or unexpected logouts.
> 
> **Overview**
> **Server-side rendering** is opt-in via `startApp({ ssr: true })`, which turns on Vite custom app mode, document SSR in `viteServer`, client + SSR builds (`ssr-manifest`, `.modelence/build/ssr`), and a streaming pipeline (`renderToPipeableStream`, session/query state scripts, CSS collection, in-process `callMethod` during render).
> 
> **Client bootstrap** changes: `renderApp` hydrates when `__MODELENCE_STATE__` is present, optional `router` + location for route parity, `ModelenceQueryProvider` / dehydration unless the app already connected a QueryClient; `AppProvider` skips the loading shell when SSR markup exists and can **reconcile** anonymous SSR sessions against a localStorage token.
> 
> **Auth for SSR**: httpOnly `authToken` cookies (`sameSite: lax`, refresh on login and logged-in `_system.session.init` only); method **context** now uses nullable `req`/`res` for in-process and live-query paths.
> 
> **React Query** helpers (`modelenceQuery`, `modelenceLiveQuery`, `modelenceMutation`, provider/connect APIs) ship from `modelence/client` with `@tanstack/react-query` as a **peer dependency**; docs nav updated. Package version **0.20.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76d71af6cc0b500100b8e698fae00f78ffd958fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


artahian (commented):
As discussed, my main comment is the backwards compatibility. It sounds like we could still keep it, but would be nice to verify.

packages/modelence/src/client/query.ts
```diff
@@ -0,0 +1,173 @@
...
+let queryClientRef: QueryClient | null = null;
+let cacheUnsubscribe: (() => void) | null = null;
+const subscriptions = new Map<string, Subscription>();
+
+/**
+ * Connects a `QueryClient` to Modelence's live-query layer. Auto-called by
+ * `<ModelenceQueryProvider>`; only call manually if you mount your own
+ * `<QueryClientProvider>`.
+ */
+export function connectModelenceQueryClient(queryClient: QueryClient) {
```
artahian, line 31: @omegascorp why do we need to add anything about live queries? Isn't it supposed to just not support SSR?

artahian (commented):
At the high level this seems too complicated, couldn't we just do what Meteor does by running a headless browser, rendering the page and then using the output to render? It would work for any scenario without having to adjust everything one by one.

packages/modelence/src/client/query.ts
omegascorp: Answer to the question above - rendering everything in a headless browser will increase latency a lot (200–2000ms) + will use more resources (CPU/Memory)

omegascorp:
@artahian @eduard-piliposyan there was an issue with ssr of homepage, the rest of the pages worked as expected. The issue is resolved the demo his here https://webappbench.com/

packages/modelence/src/app/index.ts
```diff
@@ -81,7 +83,11 @@ export async function startApp({
   auth = {},
   security = {},
   websocket = {},
+  ssr = false,
 }: AppOptions) {
+  if (ssr && server === viteServer) {
+    viteServer.enableSsr();
+  }
```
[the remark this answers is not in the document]
omegascorp: This is super edge case it's fine

packages/modelence/src/auth/session.ts
```diff
@@ -100,14 +100,31 @@ export function setAuthTokenCookie(res: Response, authToken: string) {
     secure: process.env.NODE_ENV === 'production',
     sameSite: 'strict',
     path: '/',
+    maxAge: time.days(7),
+  });
```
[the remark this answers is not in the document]
omegascorp: I don't think we care about HTTP

packages/modelence/src/bin/build.ts
```diff
@@ -13,7 +13,13 @@ async function buildClient() {
     return;
   }
 
+  // Always build the SSR bundle and emit `ssr-manifest.json`. Whether the
+  // runtime actually uses them is decided by `startApp({ ssr: true })`.
+  // Building unconditionally avoids the build-vs-runtime config mismatch
+  // class of bugs (e.g. "I enabled SSR but the deployed app serves a CSR
+  // shell because the build was missing the SSR artifact").
   await buildVite();
+  await buildViteSsr();
```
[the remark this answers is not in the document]
omegascorp: Addressed

eduard-piliposyan (approved):

packages/modelence/src/methods/types.ts
```diff
@@ -24,11 +24,10 @@ export type Context = {
...
   connectionInfo: ConnectionInfo;
-};
-
-export type HttpContext = Context & {
-  req: Request;
-  res: Response;
+  /** Express request. `null` for in-process invocations (e.g. background jobs). */
+  req?: Request | null;
+  /** Express response. Used by auth handlers to set/clear cookies. */
+  res: Response | null;
```
Abhijeet1005, line 30: Folding `req`/`res` into `Context` (and retiring the old `Context | HttpContext` TODO) is a reasonable cleanup, but two things:

**Optionality is asymmetric.** `req?: Request | null` is *optional and* nullable, while `res: Response | null` is a *required* key that's nullable — so every `Context` constructor must supply `res` (even as `null`) yet may omit `req`. I'd make them consistent (both optional-and-nullable, or both required-and-nullable) so call sites aren't surprised.

**This collides with #284.** That open PR imports `HttpContext` from this module and does `'req' in context ? context.req?.cookies?.[...] : undefined`. Removing `HttpContext` breaks that import, and since `req`/`res` become always-present keys here, the `'req' in context` narrowing there changes meaning. Whichever of #248/#284 merges second will need a fixup — worth coordinating the merge order.

packages/modelence/src/viteServer.ts
```diff
@@ -13,16 +13,57 @@ import path from 'path';
 import fs from 'fs';
 import express from 'express';
 import type { AppServer, AppServerInitOptions, ExpressMiddleware } from './types';
+import { escapeJsonForScript } from './ssr/escapeJsonForScript';
+
+const CLIENT_BUILD_DIR = './.modelence/build/client'.replace(/\\/g, '/');
+const SSR_BUILD_DIR = './.modelence/build/ssr'.replace(/\\/g, '/');
+// Resolved relative to Vite's `root` (./src/client).
+const SSR_ENTRY_VIRTUAL_PATH = '/index.tsx';
+const ROOT_PLACEHOLDER_REGEX = /<div id="root">\s*<\/div>/;
```
Abhijeet1005: This is the one recent Bugbot finding still live at head, and it's worth addressing. `/<div id="root">\s*<\/div>/` only matches a bare, double-quoted, attribute-less root div — a template with `<div id="root" class="app">`, single quotes, or extra intra-tag whitespace won't match and `splitTemplateAtRoot` throws.

The failure mode is the subtle part: that throw is caught by `handler()`'s try/catch (~L95–112), which falls back to `serveStaticShell` — so SSR doesn't 500, it **silently degrades to client-only rendering** with no error surfaced. A user who just adds a class to their root div would quietly lose SSR and likely never notice.

Two suggestions: (1) make the match tolerant, e.g. `/<div\s+id=["']root["'][^>]*>\s*<\/div>/` (or split on the located element rather than an exact string); and (2) log when the placeholder isn't found, so a silent SSR-disable becomes debuggable.

Abhijeet1005 (commented):
Took an independent pass at this against the current head (`6c566cf`), focused on the **shared / non-SSR code paths** (where an SSR PR is most likely to regress existing users) and on verifying the recent Bugbot findings against head rather than taking them at face value.

**Bot findings — status at head (most are already addressed 👍):**
- *SSR transport breaks `callMethod` globally* — **fixed.** Bot reviewed `034b62f` (where the no-context branch threw); at head it falls back to `defaultCallMethodTransport`, and the TSDoc documents exactly that case.
- *Client router omits SSR location* — **fixed.** `renderApp` now passes `location` to the router.
- *Extra query provider breaks live queries* — **fixed.** Now guarded by `hasConnectedQueryClient()`.
- *Async catch-all risks unhandled rejections* (earlier) — **fixed.** `app.all('*')` now does `Promise.resolve(server.handler(...)).catch(next)`.
- *Special `$` patterns corrupt template replacement* (earlier) — **fixed.** Assembly switched to `res.write()` streaming + `slice()`; no string-replacement remains.
- *SSR root placeholder too strict* — **still live** (see inline; it silently disables SSR rather than erroring).

So the wall of "Medium" bot comments is mostly stale — nice turnaround on the fixes.

**Worth a look (inline):**
1. **Cross-PR collision with #284.** This PR removes `HttpContext` and folds `req`/`res` into `Context`; #284 (server-side reset-token exchange, currently open and under review) imports `HttpContext` and narrows with `'req' in context`. Whichever lands second won't compile / changes meaning — worth sequencing the two deliberately.
2. The new `Context.req`/`res` optionality is asymmetric.
3. The root-placeholder regex degrades SSR silently.

Submitting as a comment — nothing here is a hard blocker, and given how fast the bot findings were addressed this is in good shape.

packages/modelence/src/methods/types.ts
omegascorp: @Abhijeet1005 thank you. I will align the other PR with this change once merged. Your suggestion to make the req/res non-optional has been applied.

packages/modelence/src/viteServer.ts
omegascorp: Thank you, the suggestion is applied

PR #263, opened 2026-05-13
feat: improved rate limits format in auth config

omegascorp:
Old format
```
rateLimits: {
  signup: { perIp15Minutes: 10, perIpPerDay: 30 },
}
```
New format:
```
rateLimits: {
  signup: [
    { type: 'ip', window: time.minutes(15), limit: 10 },
    { type: 'ip', window: time.days(1),    limit: 30 },
  ],
}
```

---

> [!NOTE]
> **Medium Risk**
> Changes the public `auth.rateLimits` configuration shape and the merge logic that determines effective auth endpoint throttling; mistakes could weaken or overly restrict protections. Back-compat is kept via deprecated legacy object support, reducing immediate breakage risk.
> 
> **Overview**
> Updates `auth.rateLimits` to accept **arrays of `{ type, window, limit }` rules** per auth bucket instead of the legacy `{ perIp15Minutes, ... }` object shape, and documents merge semantics (override by `(type, window)` while preserving other defaults).
> 
> `buildAuthRateLimits` is updated to merge array rules into defaults, allow adding new windows, and still accept the deprecated legacy object form (emitting a one-time `console.warn`); tests are expanded/updated accordingly. Exports are adjusted to surface the new `AuthRateLimitOverride` type, and the package version is bumped to `0.17.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec33e7314e50197f3503c8b595aeb65896b78904. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


artahian (commented):
Inconsistency worth flagging

The two doc comments on this branch contradict each other:

AuthRateLimitsConfig (lines 62‑84 of authConfig.ts) correctly says: "when provided, the array fully replaces the built-in defaults for that bucket."

The rateLimits?: field on AuthConfig (lines 201‑205 of authConfig.ts) says: "Only the fields you specify are overridden; all others keep their defaults." — which is true at the bucket granularity but misleading at the window granularity, since people will read "fields" as individual windows.


---

Also the high level question is how do we prevent the consumer from accidentally removing default rate limits? Since we override everything, even just adding one rate limit clears everything else for that bucket, which doesn't seem natural. Maybe we can merge it with the defaults, and override only the ones with matching bucket AND type AND window ?

omegascorp:
@artahian I was thinking about merging yesterday, but thought that it might not be required for LLM - it can duplicate the rules. Updating now

artahian (approved):

PR #271, opened 2026-05-19
chore: migrate to vitest

omegascorp:
* migrated `jest` tests to a modern alternative `vitest`

*Reasoning:*
1. There are a lot of experimental and unstable features that we were using from Jest, and Vitest supports them natively 
2. Vitest supports typescript natively (jest requires `ts-jest`)
3. Vitest is faster than Jest; it's not that important for us, but it will be a much better experience for our clients if we integrate it into the sitebuilder (for our own tests, it's 3x difference)

jest
```
Test Suites: 85 passed, 85 total
Tests:       698 passed, 698 total
Snapshots:   0 total
Time:        11.418 s
Ran all test suites.
```

vitest
```
Test Files  85 passed (85)
      Tests  698 passed (698)
   Start at  11:09:40
   Duration  3.64s (transform 2.08s, setup 0ms, collect 6.50s, tests 1.34s, environment 10ms, prepare 3.97s)
```


eduard-piliposyan (approved):
I don't have any objections on moving from jest to vitest

PR #275, opened 2026-05-21
feat: add create method to store

omegascorp:
* add create method to store, which returns the inserted document

---

> [!NOTE]
> **Medium Risk**
> New write path mutates documents via Zod internals before insert; behavior differs from insertOne and depends on default-resolution correctness for nested/optional fields.
> 
> **Overview**
> Adds **`Store.create`**, which inserts a document and returns the full record (with generated **`_id`**, schema defaults, and document helper methods) instead of only insert metadata like **`insertOne`**.
> 
> A new **`applyDefaultsToModelSchema`** helper fills missing fields from Zod **`.default()`** values (including nested objects and arrays) before the insert, so persisted data and the returned object match.
> 
> Coverage includes a store test that top-level and nested defaults are applied on both the insert payload and the **`create`** return value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32b923cf46a61f2fdcdcb5fe1311833aa77b0a8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/modelence/src/data/store.ts
```diff
@@ -917,6 +917,38 @@ export class Store<
...
+   * @returns The inserted document with `_id` populated and methods applied
+   *
+   * @example
+   * ```ts
+   * const todo = await dbTodos.create({ title: 'Buy milk', completed: false });
+   * console.log(todo._id);    // ObjectId
+   * console.log(todo.title);  // 'Buy milk'
+   * ```
+   */
+  async create(
```
eduard-piliposyan, line 938: The functionality looks good to me!
Is "create" a standard name for such a method?

eduard-piliposyan (approved):

packages/modelence/src/data/store.ts
omegascorp: I've copied it from `mongoose` both the logic and the name; MongoDB driver does not have a similar method

PR #279, opened 2026-06-03
fix: add cron jobs to _modelenceCronJobs store

AtulYadav25:
Added a new `registerNewCronJobs()` function that initializes cron job records in the MongoDB store when the application starts.

## Problem
Previously, when modelence app is started and `startCronJobs()` executed from `modelence/src/cron/jobs.ts`, it attempted to fetch cron job records from the database. 

Screenshot (752)

But, these records were never inserted into the database, so these resulted the fetch operation to fail silently and preventing proper initialization of job scheduling timestamps.

Before:

Screenshot (743)

## Solution
Added `registerNewCronJobs()` which:
- Queries the database for existing cron job records
- Inserts only new jobs that haven't been registered yet
- Prevents duplicate records with a database unique index on `alias`

The function is called during `startApp()` initialization after MongoDB connects.

## Testing
- [x] Verified new cron jobs are inserted on first run
Screenshot (745)

- [x] Verified existing jobs are not duplicated on multiple runs

---

> [!NOTE]
> **Medium Risk**
> Touches cron scheduling and Mongo persistence; wrong upsert or scheduling logic could duplicate runs or skip jobs in multi-instance deployments.
> 
> **Overview**
> **Cron job persistence** now uses **`upsertOne`** instead of **`updateOne`** when a job starts, setting **`lastStartDate`** and inserting **`alias`** on first run via **`$setOnInsert`**, so `_modelenceCronJobs` rows are created on execution rather than requiring a separate registration step.
> 
> **`startCronJobs`** bails out early when **`getMongodbUri()`** is empty (logs and does not fetch or start the interval). After loading existing records, any job still missing **`scheduledRunTs`** is scheduled from **`now`**, so first runs work even when the DB has no row yet.
> 
> Tests mock **`getMongodbUri`**, expect **`upsertOne`**, and cover the no-URI skip and **`lastStartDate`**-based scheduling paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0e4dfd9099bfbedbbf110ab252c8e657da36e8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AtulYadav25:
@omegascorp PR ready for review

omegascorp:
Thank you @AtulYadav25 , I'm checking PR, please remove `pnpm-lock.yaml`

AtulYadav25:
After this update:

- The cron jobs are added to _modelenceCronJobs collection.
- Use AcquireLock to allow only one instance register new cron jobs in _modelenceCronJobs .

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,6 +57,47 @@ export function defineCronJob(
...
+ * This prevents duplicate entries and maintains an audit trail of when each
+ * cron job was first scheduled.
+ */
+export async function registerNewCronJobs() {
+  const aliasList = Object.keys(cronJobs);
+  if (aliasList.length === 0) {
+    return;
+  }
+
+  const hasLock = await acquireLock('cron-registration', {
```
omegascorp: I think we need to reuse the existing lock; having a second lock is not necessary, the call of registerNewCronJobs is already guarded by the lock

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,6 +57,47 @@ export function defineCronJob(
...
+/**
+ * Registers newly defined cron jobs in the database (If MongoDB Client Connected).
+ *
+ * This function initializes the database registry for all cron jobs defined
+ * in the application. It compares the application's cron job definitions against
+ * the database records and inserts only those that haven't been registered yet.
+ * This prevents duplicate entries and maintains an audit trail of when each
+ * cron job was first scheduled.
+ */
+export async function registerNewCronJobs() {
```
omegascorp: we need to cover the new function with tests

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,6 +57,47 @@ export function defineCronJob(
...
+    lockDuration: time.seconds(10),
+    successfulLockCacheDuration: time.seconds(5),
+    failedLockCacheDuration: time.seconds(2),
+  });
+  if (!hasLock) {
+    return;
+  }
+
+  const aliasSelector = { alias: { $in: aliasList } };
+  const existingCronJobs = await cronJobsCollection.fetch(aliasSelector);
```
omegascorp: We need to make sure to add proper error handling here. What happens if one of those calls throws an error?

omegascorp (commented):
Thank you @AtulYadav25, I've added some comments

packages/modelence/src/cron/jobs.ts
AtulYadav25: I added `try/catch` in the function `registerNewCronJobs()`

```ts
try {
    const aliasSelector = { alias: { $in: aliasList } };
    const existingCronJobs = await cronJobsCollection.fetch(aliasSelector);
    const existingCronJobAliases = new Set(existingCronJobs.map((job) => job.alias));

    const insertItems = Object.values(cronJobs)
      .filter((job) => !existingCronJobAliases.has(job.alias))
      .map((job) => ({
        alias: job.alias,
      }));
    if (insertItems.length > 0) {
      await cronJobsCollection.insertMany(insertItems);
    }
  } catch (error) {
    await releaseLock('cron-registration'); // Release lock so other instances (or a restart) can retry this .
    throw error;
  }
```

Errors here will abort app startup.

packages/modelence/src/cron/jobs.ts
AtulYadav25: I might be wrong 😅, but I didn't find any existing guard lock on call to `registerNewCronJobs`

In `createIndexesAndMigrationsWithLock`, I can see it holds a lock until that background index task completes, so if you meant that I should reuse this `migration` lock, I think I can move `registerNewCronJobs` inside `createIndexesAndMigrationsWithLock`

This aligns with your comment and I will remove the use of separate lock from `registerNewCronJobs`

AtulYadav25:
Summary:: @omegascorp 

With this commit, `registerNewCronJobs()` function is not using any extra lock, instead it is handled by the instance which holds the `'migration'` lock.

Other instances now wait in `startCronJobs()` until all cron jobs have been registered.

And `'_modelenceCronJobs'` has been changed to use `indexCreationMode: 'blocking'`.

omegascorp:
@AtulYadav25, sorry for the delay. I finally managed to take a look. Please check the bot's findings below the important part are issues number 1 and 2

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,13 +58,62 @@ export function defineCronJob(
...
+    const registeredAliases = new Set(records.map((r) => r.alias));
+    if (aliasList.every((alias) => registeredAliases.has(alias))) {
+      return;
+    }
+    await new Promise((resolve) => setTimeout(resolve, pollInterval));
+  }
+}
+
 export async function startCronJobs() {
   if (cronJobsInterval) {
```
modelence-ops, line 62: **This blocking wait runs on paths where registration never happens.** `startCronJobs()` is called unconditionally at the end of `startApp`, including:

- **No MongoDB URI:** `connect()` and `registerNewCronJobs()` are skipped entirely, so this polls a never-populated store for the full 30s before proceeding (regression — previously this path did a single `fetch` and returned immediately).
- **Instances that don't hold the migrations lock:** waiting for the lock-holder to register is the intended behavior here and is a real improvement — the only risk is the 30s ceiling if the lock-holder is slow or dies.

Consider skipping the wait when there's no Mongo URI (e.g. only wait when registration is actually expected to run).

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,13 +58,62 @@ export function defineCronJob(
...
+      .map((job) => ({
+        alias: job.alias,
+      }));
+
+    if (insertItems.length > 0) {
+      await cronJobsCollection.insertMany(insertItems);
+    }
+  } catch (error) {
+    throw error;
+  }
```
modelence-ops: **The timeout expires silently.** When the deadline passes without all aliases registered, this returns `void` with no log or error. `startCronJobs` then proceeds, the follow-up `fetch` returns fewer records, and those jobs silently fall back to `scheduledRunTs = now`. In production this failure mode would be invisible.

Suggest logging a warning when the deadline is hit, listing the aliases still missing — e.g.:

```ts
console.warn(
  `Timed out after ${timeout}ms waiting for cron jobs to be registered: ${missing.join(', ')}`
);
```

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,13 +58,62 @@ export function defineCronJob(
...
+    const existingCronJobs = await cronJobsCollection.fetch(aliasSelector);
+    const existingCronJobAliases = new Set(existingCronJobs.map((job) => job.alias));
+
+    const insertItems = Object.values(cronJobs)
+      .filter((job) => !existingCronJobAliases.has(job.alias))
+      .map((job) => ({
+        alias: job.alias,
+      }));
+
+    if (insertItems.length > 0) {
```
modelence-ops: The `try`/`catch` here is a no-op — it only re-throws, which is identical to not catching at all. You can drop the `try`/`catch` wrapper entirely and let the body run directly.

Separately, since `registerNewCronJobs` only runs under the migrations lock, the read-then-`insertMany` is safe from concurrent inserts — worth a one-line comment noting it relies on that lock for mutual exclusion, since an ordered `insertMany` would abort on a duplicate-key error if that assumption ever changed.

packages/modelence/src/app/index.ts
```diff
@@ -252,6 +257,9 @@ async function createIndexesAndMigrationsWithLock(
     for (const store of blockingStores) {
       await createStoreIndexes(store, 'full');
     }
+    // registerNewCronJobs runs here — after the unique alias index is guaranteed
+    // to exist (cronJobsCollection is now blocking!)
+    await registerNewCronJobs();
```
modelence-ops: Design question on severity: because this `await` is inside the `try` block, a transient DB hiccup during cron registration now aborts the **entire** `startApp` (and releases the migrations lock, so other instances may proceed without migrations finished).

That's inconsistent with index creation right above/below it — `createStoreIndexes` catches failures via `warnIndexCreationFailure` and startup continues. Is crashing the whole app the intended severity for cron registration specifically? If not, consider catching + warning here to match the "continue startup on index failure" philosophy.

modelence-ops (commented):
Thanks for this — the core fix is correct and well-targeted: cron rows were never inserted, so `startCronJobs` fetched nothing and timestamps never persisted. Registering rows up front under the migrations lock and switching to a blocking unique index is a sound approach, and the test coverage is solid.

A few things worth addressing before merge (left as inline comments):

- **`waitForCronJobsRegistered` blocks paths where registration never runs** (no-MongoDB-URI especially) — a 30s stall on what was previously instant.
- **The 30s timeout expires silently** — no log when aliases are still unregistered, making the failure mode invisible in prod.
- **Crash-startup severity** — a transient DB error during `registerNewCronJobs()` now aborts all of `startApp` and releases the migrations lock, inconsistent with how index-creation failures are tolerated right next to it. Worth confirming that's intended.
- **Nit:** the `try`/`catch` in `registerNewCronJobs` only re-throws — it can be removed.

Nice touches: `updateOne` → `upsertOne` with `$setOnInsert` is the right defensive move, the blocking-then-background index ordering is correctly reasoned, and the `vi.useFakeTimers({ toFake: ['setTimeout'] })` test changes correctly keep the new poll loop from hanging tests.

(Findings #1 and #2 are both new in this PR — `waitForCronJobsRegistered` doesn't exist on `main`.)

AtulYadav25:
# Summary: Database Stores Cron Jobs and Resilient Cluster Startup

This PR stores the cron jobs in mongodb and their history to coordinate execution across multiple application instances, along with indexing, startup locking, and polling optimizations.

## 🚀 Key Features
### Database-Backed Cron Jobs `(_modelenceCronJobs)`
- MongoDB has store `_modelenceCronJobs` to keep track of cron job history (alias and lastStartDate).
- When starting up, cron jobs load their schedules based on their last execution date in the database: 

```ts
scheduledRunTs = lastStartDate + interval
```
This prevents double executions across restarts.
- New or unregistered cron jobs fallback to running immediately on startup of the lock holder instance.

### Distributed Registration Lock
- Uses the `migration` lock to ensure only one instance (the lock-holder) executes `registerNewCronJobs()` at startup. This prevents duplicate insertion attempts of cron jobs and race conditions.

## 🛠️ Improvements & Bug Fixes

### Non-Blocking Startup when MongoDB is Offline/Not Configured

- Gated database-related startup operations on `getMongodbUri()`. If MongoDB is not configured, instances bypass the polling wait entirely, preventing a 30-second startup hang of waiting for cron jobs to register and scheduling all jobs immediately.

### Registration Status Document for Early Polling Termination

- The lock-holder creates a document in `_modelenceCronJobs` with alias as  `_registration_status` in the database marked as `status: running`. Upon success or failure, it updates this status to `success` or `failed`.
- Secondary instances poll the status document and exit the `waitForCronJobsRegistered` loop immediately when registration is complete, rather than waiting out the full 30-second timeout.

### Stale Session Token Validation (Race Condition Fix)

- To prevent secondary instances from matching a stale `_registration_status` document left in the database from a previous boot/run, a `sessionToken` field was added.
- The sessionToken is bound to the current migrations lock's `instanceId`. Secondary instances only respect the 'success' or 'failed' status if it matches the current session token.

### Error Handling
- Wrapped the lock-holder's `registerNewCronJobs()` call in a `try/catch` block that logs a warning and continues startup instead of crashing the whole application on transient database hiccups.
- Added `console.warn` notifications when registration waits time out, indicating which aliases failed to register.

@omegascorp I hope this summary is helpful

packages/modelence/src/cron/jobs.ts
```diff
@@ -29,6 +34,10 @@ export function defineCronJob(
     handler,
   }: CronJobInputParams
 ) {
+  if (alias === '_registration_status') {
```
omegascorp: This feels like a hack; the status seems to be part of the lock for some reason we store it here, maybe it will be better to modify the lock collection to carry status

packages/modelence/src/cron/jobs.ts
```diff
@@ -57,26 +65,155 @@ export function defineCronJob(
...
+        { $set: { status: CRON_REGISTRATION_STATUS.FAILED } }
+      );
+    } catch {
+      // Swallowed to prevent hiding original error
+    }
+    throw error;
+  }
+}
+
+async function waitForCronJobsRegistered(aliasList: string[], timeout: number): Promise<void> {
```
modelence-ops: Is the `waitForCronJobsRegistered` startup barrier actually needed? Tracing it through, I don't think it changes any observable behavior:

- **New aliases:** `registerNewCronJobs` inserts `{ alias }` with no `lastStartDate`. So in `startCronJobs`, `record.lastStartDate ? … : now` evaluates to `now`. A brand-new job schedules at `now` whether or not the barrier waited for its row to appear — identical to the no-record fallback right below (`if (!job.state.scheduledRunTs) job.state.scheduledRunTs = now`).
- **Existing aliases:** the row (with `lastStartDate`) already exists from a prior boot, so the follow-up `fetch` sees it with no wait needed.
- **Double-runs:** already prevented at execution time by `acquireLock('cron', …)` in `tickCronJobs` — only one instance runs a given tick regardless of scheduling skew across replicas.

If that holds, the barrier plus the `cron_registered` / `cron_registration_failed` status piggybacked on the migrations lock (and most of the race surface that drove the last ~15 Bugbot findings) can be dropped with no behavior change — leaving the actual fix intact: register rows under the migrations lock, blocking unique `alias` index, and `upsert`-on-run. That also resolves @omegascorp's earlier "feels like a hack" concern about storing status on the lock doc.

Am I missing a case where a registered-but-not-yet-run job needs a DB-sourced `scheduledRunTs` at startup? If there's a planned direction (e.g. registration seeding an initial `lastStartDate`), disregard — but then it's worth a comment spelling that out, because as written the barrier's only effect is a 30s startup cost on the paths where it does fire.

packages/modelence/src/lock/db.ts
```diff
@@ -12,7 +12,8 @@ export const locksCollection = new Store('_modelenceLocks', {
     _id: schema.string(), // unique identifier for the lock, used as the primary key
     instanceId: schema.string(),
     acquiredAt: schema.date(),
-
+    status: schema.string(), // lock lifecycle: 'active' | 'released' | 'expired'.
```
modelence-ops: Two nits on this new field:

1. **Make it optional.** It's declared required (`schema.string()`) but is only ever populated by `acquireLock` on acquire/takeover. Any in-flight `_modelenceLocks` document written before this change — or by an older instance mid rolling-deploy — won't carry `status`, so a strict read could trip validation. `schema.string().optional()` is safer.
2. **Comment doesn't match the values.** It lists `'active' | 'released' | 'expired'`, none of which the code uses. Actual values are `'acquired'` (helpers.ts) and `'cron_registered'` / `'cron_registration_failed'` (cron/jobs.ts). Worth aligning the comment with reality to avoid confusing the next reader.

modelence-ops (commented):
Following up on my earlier review (posting at Hayk's request). Two items to weigh before this merges — neither disputes the core fix, which is sound:

1. An architectural-complexity question on the `waitForCronJobsRegistered` startup barrier (inline on `cron/jobs.ts`).
2. A minor schema/back-compat nit on the new `_modelenceLocks.status` field (inline on `lock/db.ts`).

AtulYadav25:
Thank you @modelence-ops & @omegascorp 

Good point `waitForCronJobsRegistered` function was redundant. I initially planned for problem of limiting the number of cron jobs running concurrently, but this can be done more cleanly without the `waitForCronJobsRegistered` function.

The PR now has cleaner code:
- It has only the `registerNewCronJobs` function, which runs only when it has `_modelenceCronJobs` store
- In `startCronJobs`, the function queries the database only when MongoDB is configured
- Added test cases for all the changes.

modelence-ops (commented):
Re-reviewed at `bb7acdd`. This is exactly the simplification I was hoping for — dropping `waitForCronJobsRegistered` and the status-on-lock / `sessionToken` / `_registration_status` machinery collapses the whole race surface that drove most of the Bugbot findings, and (as traced in my earlier comment) it does so with no observable behavior change:

- New aliases insert `{ alias }` with no `lastStartDate`, so `startCronJobs` schedules them at `now` — same as the no-record fallback.
- Existing aliases already carry `lastStartDate` from a prior boot; the `getMongodbUri()`-gated fetch picks them up.
- Cross-instance double-runs stay prevented by `acquireLock('cron', …)` in `tickCronJobs`, independent of scheduling skew.

That also resolves @omegascorp's "feels like a hack" concern about storing status on the lock doc, and my `_modelenceLocks.status` back-compat nit is moot now that `lock/db.ts` is untouched. The `forEach(async …)` → `for…of await` change in `tickCronJobs` is a real correctness win (the old form fired unawaited handlers and swallowed rejections), and the new coverage on `registerNewCronJobs` / `isDuplicateKeyError` / the no-Mongo path is solid. I've resolved my two prior threads.

Two minor things, neither blocking:

1. The comment above `insertMany` says an "ordered insertMany would abort on a duplicate-key error," but the call passes `{ ordered: false }`. With `ordered: false` the driver keeps going and surfaces dups via `writeErrors` (which `isDuplicateKeyError` now handles) — worth aligning the comment with the code.
2. Bugbot's "cron writes without unique index" edge (blocking index build fails → registration skipped, but `runCronJob` still upserts): in practice the single `acquireLock('cron')` holder serializes those upserts, so duplicate alias docs shouldn't actually occur, and it self-heals once the index lands. Fine to leave as-is — just noting it's a conscious call.

Core fix is sound and the architecture is now where it should be. Leaving the merge call to @omegascorp / @eduard-piliposyan. Thanks @AtulYadav25 for pushing this through the cleanup.

artahian (dismissed):
@AtulYadav25 the real bug here is the updateOne no-op in runCronJob - your upsertOne + $setOnInsert change should fix it completely. I don't think `registerNewCronJobs` is needed: a row without lastStartDate behaves the same as no row, so all the registration + blocking index + duplicate-key machinery has zero effect.

Can we please revert the tick loop change to sequential await - a slow job would block all others, and it's unrelated to this fix.

As for the getMongodbUri() guard, we don't really want to support a use case of running apps without a db, it's just not a Modelence priority. But if that's somehow the case, cron jobs should still not be supported, they require locks and persistence to properly run - let's switch to simply not start any cron jobs if there's no db.

AtulYadav25:
Thanks @artahian  ,catching this
I was again overengineering this way more than it needed, `upsertOne` fix alone handles everything 

I pulled out all the extra stuff

Tested locally in a Modelence app and confirmed jobs get added to `_modelenceCronJobs`.

I pushed the clean version commit

omegascorp (approved):
@AtulYadav25 it finally looks.

PR #281, opened 2026-06-15
MOD-432: fix: email verification url

omegascorp:
* Fixed email verification URL
* Add error codes to responses
* `_system.user.auth.email.verification` was ignored before 

---

> [!NOTE]
> **Medium Risk**
> Changes login gating defaults and trusts forwarded headers for base URLs; misconfigured proxies or the new verification default could block logins or affect link generation until `_system.site.url` or config is set correctly.
> 
> **Overview**
> Fixes **email verification links** and **public base URLs** when the app runs behind a reverse proxy, and makes **email verification at login** honor system config with **machine-readable error codes** for clients.
> 
> **Proxy / URL handling:** `connectionInfo.baseUrl` now comes from `getRequestBaseUrl`, which prefers `X-Forwarded-Host` and `X-Forwarded-Proto` (first value in comma-separated lists) instead of the internal `Host` / `req.protocol`. Verification emails build links with `_system.site.url` when set, otherwise the request-derived `baseUrl`.
> 
> **Login verification:** Password login no longer blocks unverified users solely because an email provider exists. `isEmailVerificationRequired()` requires a provider **and** `_system.user.auth.email.verification` (schema default changed from `false` to `true`). Unverified login when required throws `AuthError` with code `EMAIL_NOT_VERIFIED`.
> 
> **Error codes:** `ModelenceError` / `AuthError` accept an optional `code`; method errors set `X-Modelence-Error-Code` when present. The client `MethodError` reads that header so UIs can branch without parsing message text.
> 
> Package version bumped to **0.19.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3af6f333b710ba19fc75301868a0ef4415e12e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/user.ts
+++ b/packages/modelence/src/auth/user.ts
@@ -161,7 +161,7 @@ export default new Module('_system.user', {
     'auth.email.verification': {
       type: 'boolean',
       isPublic: true,
-      default: false,
+      default: true,
     },
     'auth.google.enabled': {
       type: 'boolean',
--- a/packages/modelence/package-lock.json
+++ b/packages/modelence/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "modelence",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "modelence",
-      "version": "0.19.0",
+      "version": "0.19.1",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@socket.io/mongo-adapter": "^0.4.0",
--- a/packages/modelence/package.json
+++ b/packages/modelence/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.19.0",
+  "version": "0.19.0-email.dev.0",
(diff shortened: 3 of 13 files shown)


packages/modelence/src/auth/user.ts
```diff
@@ -161,7 +161,7 @@ export default new Module('_system.user', {
     'auth.email.verification': {
       type: 'boolean',
       isPublic: true,
-      default: false,
+      default: true,
```
eduard-piliposyan, line 164: Is this intentional? Do we want email verification be on for all new created modelence sandboxes?

eduard-piliposyan (approved):
All looks good to me except that we changed the default behavior of email verification. Do we really need that?

packages/modelence/src/auth/user.ts
omegascorp: Yes, this aligned the real behavior with the property values. We didn't use this property before, so the agent was assuming it was disabled, but it was not. (I had an issue because of this - agent was trying to sign in right after sign-up).

Why it worked before: we didn't have default email integration, so we were considering all emails as verified by default.

PR #282, opened 2026-06-19
Update: Added User schema extended fields to auth user object

AtulYadav25:
```ts
import { schema, dbUsers } from 'modelence/server';

export const extendedDbUsers = dbUsers.extend({
  schema: {
    companyId: schema.string().optional(),
  },
  indexes: [
    { key: { companyId: 1 } },
  ],
});
```

When used `extend` to extend schema of system users collection with custom fields, the `authenticate()` function in `modelence/src/auth/index.ts` returned user object by manually listing a fixed set of fields (firstName, lastName, avatarUrl, etc.) instead of relying on the full document.

So any custom field added through schema extension was present in the database document (userDoc) but not in the user object returned to the app via this `authenticate()`

PR:
`authenticate()` now spreads the full safeUser object (the userDoc with only the unsafe _id and authMethods fields stripped) as the base of the returned user.

Ensures any field present on the document, including ones added via schema extension, is automatically forwarded.

## For
```ts
 async getChats(args:any, { user }:{user: any}) {
      console.log(user);
}
```

### Before: 
extending the schema and saving a new field (e.g. companyId) → field exists in MongoDB, but missing from the user object returned by authenticate().

Screenshot (769)

---
### After: 
same extended field is present on the user object as expected.

Screenshot (768)

---

> [!NOTE]
> **Medium Risk**
> Touches the authentication user-mapping path and exposes additional stored user fields to request context, though sensitive `authMethods` remain excluded.
> 
> **Overview**
> **`authenticate()`** now builds the returned user from the MongoDB document instead of a fixed field list. It strips **`_id`** and **`authMethods`**, spreads the remaining document (including fields added via **`dbUsers.extend`**), then applies the same **`id`**, role helpers, and optional name/avatar overrides as before.
> 
> Custom schema fields such as **`companyId`** are now available on **`user`** in method handlers without changing auth code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b6c5076457e9ef1db2f3b3d5e76be53a331cad2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/index.ts
+++ b/packages/modelence/src/auth/index.ts
@@ -16,21 +16,27 @@ export async function authenticate(
         status: { $nin: ['deleted', 'disabled'] },
       })
     : null;
+  //Destructure unsafe fields
   const user = userDoc
-    ? {
-        id: userDoc._id.toString(),
-        handle: userDoc.handle,
-        roles: userDoc.roles || [],
-        hasRole: (role: string) => (userDoc.roles || []).includes(role),
-        requireRole: (role: string) => {
-          if (!(userDoc.roles || []).includes(role)) {
-            throw new Error(`Access denied - role '${role}' required`);
-          }
-        },
-        firstName: userDoc.firstName ?? undefined,
-        lastName: userDoc.lastName ?? undefined,
-        avatarUrl: userDoc.avatarUrl ?? undefined,
-      }
+    ? (() => {
+        const { _id, authMethods, ...safeUser } = userDoc;
+
+        return {
+          ...safeUser,
+          id: userDoc._id.toString(),
+          handle: userDoc.handle,
+          roles: userDoc.roles || [],
+          hasRole: (role: string) => (userDoc.roles || []).includes(role),
+          requireRole: (role: string) => {
+            if (!(userDoc.roles || []).includes(role)) {
+              throw new Error(`Access denied - role '${role}' required`);
+            }
+          },
+          firstName: userDoc.firstName ?? undefined,
+          lastName: userDoc.lastName ?? undefined,
+          avatarUrl: userDoc.avatarUrl ?? undefined,
+        };
+      })()
     : null;
 
   const roles = user ? getDefaultAuthenticatedRoles() : getUnauthenticatedRoles();


packages/modelence/src/auth/index.ts
```diff
@@ -16,21 +16,27 @@ export async function authenticate(
...
-        },
-        firstName: userDoc.firstName ?? undefined,
-        lastName: userDoc.lastName ?? undefined,
-        avatarUrl: userDoc.avatarUrl ?? undefined,
-      }
+    ? (() => {
+        const { _id, authMethods, ...safeUser } = userDoc;
+
+        return {
+          ...safeUser,
```
omegascorp, line 25: This is quite dangerous; when the user object is extended, it might have other sensitive fields.

omegascorp (changes requested):
After this change authenticate() also forwards emails (with verification status), status, disabledAt, and deletedAt from the base schema — none of which the previous allowlist exposed — plus any future field added via dbUsers.extend(...). The whole premise of the PR is that extended fields are unpredictable, which is exactly the argument against auto-forwarding them by default: an extend({ schema: { stripeCustomerId, ssn, internalRiskScore } }) would now land all of those on user in every method handler silently. Allowlists fail closed; denylists fail open, and here authMethods is the only thing between this and leaking password hashes.

Also worth noting: UserInfo still types only the 7 known fields, so the extended fields aren't actually typed on user — the stated DX goal isn't reached type-safely while the leak happens at runtime.

Proposed direction instead: make field exposure a property of the schema, with a .private() marker that the framework strips during serialization:


authMethods: schema.object({ ... }).private(),
emails: schema.array(...).private(),
Since schema already flows through InferDocumentType and merges via extend(), one marker is honored everywhere authenticate() builds the user object — and an app that extends with ssn: schema.string().private() is protected automatically. authenticate() then becomes a generic stripPrivateFields(doc, schema) instead of a hardcoded { _id, authMethods } strip, and the original goal (extended fields on user) works and stays opt-out-able.

(Note: this cleanly covers the authenticate() path. The method-return serializer (sanitizeResult) is schema-agnostic today, so plumbing .private() through there is a larger follow-up — but authenticate() is the path this PR touches, so it's the right place to start.)

packages/modelence/src/auth/index.ts
AtulYadav25: Agree with points, didn't think about it from that angle 👏, it is risky, extend() can add anything.

I'll rework this from your `.private()` solution, make those as private on the schema, then have authenticate() strip private fields generically instead of hardcoding our user fields... 

`sanitizeResult` gap makes sense, I'll follow up if I can help after.

Will push an update once it's in.

omegascorp: Thank you @AtulYadav25

AtulYadav25: I'll continue working on this PR after your Magic Link feature PR is merged.
Our PRs touch the same context, so I think they might clash. I will pick this back up once your PR is done.

omegascorp: Ok

omegascorp:
@AtulYadav25 magic link is merged

AtulYadav25:
> @AtulYadav25 magic link is merged

Thank you 
I'll figure out to apply the .private() feat as requested

AtulYadav25:
Continued in PR #328

omegascorp:
@AtulYadav25 what's the status of this PR, should we close it?

AtulYadav25:
> @AtulYadav25 what's the status of this PR, should we close it?

I initially thought if we could merge this because the full user data is only sent server-side and not to the client. But, passing the user object directly through the API is still risky, if this is accidentally be included in an API response by a dev/AI

So as Aram suggested, we should allow only whitelisted fields instead of blacklisting fields. I’ll close this PR, and push solutions in #328

PR #283, opened 2026-06-20
feat: Allows user to use custom authentication

AtulYadav25:
Modelence's internal session functions `(createSession, setSessionUser, obtainSession, setAuthTokenCookie)`  these already handle everything needed for authentication, but they were only used internally for modelence only authentication.

## Change
Re-exports the existing session functions from `modelence/src/server.ts` so users can do their own authentication flows using Modelence's session primitives instead of creating their own session handling from scratch.

No logic changes, this is purely making existing internal functions public.

### Example Usage After this PR:
```ts
async connectWallet(args, { req, session, res }) {

  // Mock: verifying signature against walletAddress
  // (In production: verify args.signature was signed by args.walletAddress)

  // Find existing user by wallet address
  let user = await extendedDbUsers.findOne({
    walletAddress: args.walletAddress
  });

  if (!user) {
    // No existing user - register a new one
    user = await extendedDbUsers.create({
      walletAddress: args.walletAddress,
      status: "active",
      createdAt: new Date(),
      authMethods: {},
    });
  }

  // Session already exists at this point when the method is called, just attach the userId to it
  await setSessionUser(session.authToken, user._id);

  // Add the auth token in the client cookie, next time client calling any method, modelence internally checks for cookies and find this as authenticated
  await setAuthTokenCookie(res, session.authToken);

  return {
    message: "Login successful"
  };
},
```

---

> [!NOTE]
> **Medium Risk**
> No implementation changes, but exposing session/cookie APIs shifts responsibility for correct auth flows to consumers; misuse could weaken session binding or cookie handling.
> 
> **Overview**
> **Public session primitives** are now exported from `packages/modelence/src/server.ts` so apps can wire custom login flows (e.g. wallet connect) without reimplementing Modelence’s session storage and cookies.
> 
> The new exports from `./auth/session` are: `dbSessions` (`sessionsCollection`), `createSession`, `obtainSession`, `setSessionUser`, `clearSessionUser`, `invalidateAllUserSessions`, and `setAuthTokenCookie`. Behavior is unchanged—only the package entrypoint surface grows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 377166c89e483ae4c948364a90e69d1dbc845464. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/server.ts
+++ b/packages/modelence/src/server.ts
@@ -23,6 +23,17 @@ export { usersCollection as dbUsers } from './auth/db';
 export type { UserInfo, RoleDefinition } from './auth/types';
 export { deleteUser, disableUser } from './auth/deleteUser';
 
+//Auth Session
+export {
+  sessionsCollection as dbSessions,
+  createSession,
+  clearSessionUser,
+  obtainSession,
+  setSessionUser,
+  invalidateAllUserSessions,
+  setAuthTokenCookie,
+} from './auth/session';
+
 // Database
 export { schema } from './data/types';
 export { Store } from './data/store';


packages/modelence/src/server.ts
```diff
@@ -23,6 +23,17 @@ export { usersCollection as dbUsers } from './auth/db';
 export type { UserInfo, RoleDefinition } from './auth/types';
 export { deleteUser, disableUser } from './auth/deleteUser';
 
+//Auth Session
```
Abhijeet1005, line 26: Two cross-cutting points for this whole block:

**Consider trimming to the minimal surface the use case needs.** The PR's wallet example only uses `setSessionUser` + `setAuthTokenCookie` — the session already exists by the time a method runs (created in `getCallContext`). Exposing `createSession`, `obtainSession`, and `dbSessions` adds public contract you'll have to preserve, for capabilities the documented flow doesn't exercise.

**The `Session` type isn't exported.** `createSession` and `obtainSession` both return `Session` (from `./auth/types`), but it isn't re-exported here or in `types.ts`, so a consumer can't write `const s: Session = await obtainSession(...)` without importing from an internal path. If these go public, their types need to go public too — ideally via `types.ts` per the repo's convention. (Nit: `//Auth Session` → `// Auth Session`.)

packages/modelence/src/server.ts
```diff
@@ -23,6 +23,17 @@ export { usersCollection as dbUsers } from './auth/db';
...
+export {
+  sessionsCollection as dbSessions,
```
Abhijeet1005, line 28: Re-exporting the raw `sessionsCollection` as `dbSessions` is the export I'd push back on hardest. It hands consumers unrestricted read/write/delete on the internal sessions store and pins you to its current schema. The example never touches it, and the schema is already slightly out of step with runtime — `processSessionHeartbeat` writes `lastActiveDate`, which isn't in the declared `schema` — so publishing the Store as-is makes that drift part of the public contract. I'd drop `dbSessions` unless there's a concrete need the helper functions don't cover.

packages/modelence/src/server.ts
```diff
@@ -23,6 +23,17 @@ export { usersCollection as dbUsers } from './auth/db';
...
+  createSession,
```
Abhijeet1005, line 29: `createSession` (and `obtainSession`, which mints a session when none matches) still carry the in-file `// TODO: add rate-limiting and captcha handling`. Promoting them to the public API means any custom flow can create sessions with no throttling — a resource-exhaustion vector that the TODO acknowledges isn't handled yet. Since the documented flow relies on the auto-created session rather than minting its own, I'd either keep session creation internal for now or resolve the rate-limiting TODO before exposing it.

packages/modelence/src/server.ts
```diff
@@ -23,6 +23,17 @@ export { usersCollection as dbUsers } from './auth/db';
...
+  clearSessionUser,
+  obtainSession,
+  setSessionUser,
```
Abhijeet1005, line 32: Two things on `setSessionUser` as a public primitive:

1. **Silent no-op on a bad token.** It's `updateOne({ authToken }, { $set: { userId } })` with no existence check, so if `authToken` matches no session it returns normally with `matchedCount: 0` — a custom flow would think login succeeded when nothing was bound. Consider returning the update result or throwing when nothing matched.
2. **The documented example assumes `session` is always present** (`setSessionUser(session.authToken, ...)`), but a method context's `session` can be `null` — the `heartbeat` mutation in this same module guards `if (session)` precisely because "Session might not exist if there is no database/authentication setup." The example (and the eventual docs) should guard the nullable case.

Abhijeet1005 (commented):
Nice idea — letting apps drive their own auth (wallet connect, etc.) on top of Modelence's session primitives is genuinely useful, and I confirmed this is a pure additive re-export with no logic change. My notes are all about what it means to promote *these particular* internals to the **public** API surface, since that's a contract you'll have to keep.

**Main themes (details inline):**
- **Surface is broader than the use case needs.** The motivating example only calls `setSessionUser` + `setAuthTokenCookie`. `createSession`, `obtainSession`, and especially the raw `dbSessions` collection aren't used by it — exposing them couples consumers to internals you'll want to evolve, and `createSession` still has an open `// TODO: add rate-limiting and captcha handling`.
- **The public API is incomplete without its types.** `createSession`/`obtainSession` return `Session`, but `Session` isn't re-exported from `server.ts` or `types.ts`, so consumers can't annotate their code without reaching into internal paths.
- **No TSDoc on any of the seven symbols.** They'll render blank in the generated API reference (docs.modelence.com), which is unfortunate for security-sensitive primitives — a one-line doc on `setSessionUser`/`setAuthTokenCookie` stating the caller is responsible for verifying identity first would go a long way. (Tiny nit: `//Auth Session` is missing the space the surrounding comments use.)

Submitting as a comment — these are design calls for the maintainers, not blockers.

omegascorp:
Thank you @AtulYadav25, please take a look on @Abhijeet1005 comments, I will followup after that.

@Abhijeet1005 thank you for reviewing

omegascorp (approved):
@AtulYadav25 just checked the code, for some reason, I was expecting something more complex, approved

PR #284, opened 2026-06-24
MOD-442: feat: server-side token exchange for password reset

omegascorp:
Move the password-reset token out of client-rendered URLs to close a class of leak where analytics/ad scripts (e.g. gtag) read document.location.href and ship the token to third parties.

- Add GET /api/_internal/auth/reset-password landing route: validates the token, stores it in a short-TTL httpOnly+Secure+SameSite cookie, and 302-redirects to the tokenless SPA page. Reset emails link here instead of to the SPA page directly.
- resetPassword mutation reads the token from the cookie (token arg now optional, kept for backwards compat with a deprecation warning).
- Hash reset tokens at rest (SHA-256) with a legacy plaintext fallback; consume atomically via findOneAndDelete (single-use, no TOCTOU race).
- Set Referrer-Policy: no-referrer inline on the reset-password and verify-email landing redirects so the token URL can't leak via Referer.
- Fix createRouteHandler applying response.headers after res.redirect() (headers were silently dropped on redirects).

---

> [!NOTE]
> **High Risk**
> Changes core password-reset and session-invalidation behavior, cookie/CORS requirements for cross-origin SPAs, and breaks in-flight plaintext reset tokens after deploy (new tokens are hash-only).
> 
> **Overview**
> Password reset no longer puts the raw token on SPA URLs or in client JS. Reset emails link to **`GET /api/_internal/auth/reset-password`**, which validates the token, stores it in a short-lived **httpOnly** cookie, and **302**s to the configured **`redirectUrl`** without the token. **`resetPassword({ password })`** reads the token from that cookie; an optional **`token`** argument remains for legacy clients (deprecated).
> 
> **At rest and consumption:** new tokens are stored as **SHA-256** digests only (no plaintext DB match / hash-replay path). Consumption is **`findOneAndDelete` by `_id`** at commit time for single-use and concurrent-request safety. Sending reset mail **requires** a resolvable site base URL (`_system.site.url` / `MODELENCE_SITE_URL`).
> 
> **Plumbing:** client **`callMethod`** uses **`credentials: 'include'`** so the reset cookie works cross-origin (with proper CORS). **`createRouteHandler`** sets response headers **before** redirects and skips a body on redirect; reset and verify-email landings add **`Referrer-Policy: no-referrer`**. Telemetry **`redactSensitive`** scrubs tokens/passwords from method and route transaction context. Docs describe the new flow and cookie/CORS notes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6758cb1acb036b8c32317c123115f208ac204fa9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/resetPassword.ts
+++ b/packages/modelence/src/auth/resetPassword.ts
@@ -21,6 +21,18 @@ const RESET_PASSWORD_COOKIE = 'resetPasswordToken';
 // Scope the cookie to the API surface so it's only sent on the resetPassword request.
 const RESET_PASSWORD_COOKIE_PATH = '/api/_internal/';
 
+// Attributes shared by the set and clear calls. Browsers only remove a cookie
+// when the clearing Set-Cookie matches the same path/secure/sameSite/httpOnly,
+// so both sites must use these identical flags or the httpOnly cookie can linger.
+const RESET_PASSWORD_COOKIE_OPTIONS = {
+  httpOnly: true,
+  secure: process.env.NODE_ENV === 'production',
+  // `lax` so the cookie survives the email-link navigation; `strict` can drop
+  // it when the SPA is reached cross-site (some webviews).
+  sameSite: 'lax',
+  path: RESET_PASSWORD_COOKIE_PATH,
+} as const;
+
 /**
  * Looks up a reset token without consuming it. Only matches the hashed form:
  * tokens are stored as `hashToken(rawToken)`, and the caller must present the
@@ -173,12 +185,7 @@ export async function handleResetPasswordLanding(params: RouteParams): Promise<R
     }
 
     params.res.cookie(RESET_PASSWORD_COOKIE, token, {
-      httpOnly: true,
-      secure: process.env.NODE_ENV === 'production',
-      // `lax` so the cookie survives the email-link navigation; `strict` can drop
-      // it when the SPA is reached cross-site (some webviews).
-      sameSite: 'lax',
-      path: RESET_PASSWORD_COOKIE_PATH,
+      ...RESET_PASSWORD_COOKIE_OPTIONS,
       maxAge: time.hours(1),
     });
 
@@ -222,7 +229,9 @@ export async function handleResetPassword(args: Args, context: Context) {
 
   const clearCookie = () => {
     // `res` is null for in-process invocations (no response to mutate).
-    context.res?.clearCookie(RESET_PASSWORD_COOKIE, { path: RESET_PASSWORD_COOKIE_PATH });
+    // Clear with the same flags used to set it; browsers ignore a clear whose
+    // attributes don't match, which would leave the httpOnly cookie behind.
+    context.res?.clearCookie(RESET_PASSWORD_COOKIE, RESET_PASSWORD_COOKIE_OPTIONS);
   };
 
   // Look up WITHOUT consuming: if validation/hashing fails (or the user is
--- a/packages/modelence/src/auth/resetPassword.ts
+++ b/packages/modelence/src/auth/resetPassword.ts
@@ -2,7 +2,8 @@ import { z } from 'zod';
 import bcrypt from 'bcrypt';
 import { randomBytes } from 'crypto';
 
-import { Args, Context } from '@/methods/types';
+import { Args, Context, HttpContext } from '@/methods/types';
+import { RouteParams, RouteResponse } from '@/server';
 import { usersCollection, resetPasswordTokensCollection } from './db';
 import { getEmailConfig } from '@/app/emailConfig';
 import { time } from '@/time';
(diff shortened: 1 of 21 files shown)


packages/modelence/src/auth/resetPassword.ts
```diff
@@ -86,19 +135,21 @@ export async function handleSendResetPasswordToken(args: Args, { connectionInfo
...
 
-  // Build reset URL
+  // Point the email at the server-side landing route rather than the SPA page.
+  // The landing route validates the token, moves it into an httpOnly cookie, and
+  // redirects to the tokenless SPA page — so the raw token never reaches a
+  // client-rendered URL where analytics/ad scripts could read it.
   const baseUrl = (getConfig('_system.site.url') as string | undefined) || connectionInfo?.baseUrl;
-  const resetPasswordUrl = resolveUrl(baseUrl!, getEmailConfig().passwordReset?.redirectUrl);
-  const resetUrl = `${resetPasswordUrl}?token=${resetToken}`;
+  const resetUrl = `${baseUrl}/api/_internal/auth/reset-password?token=${resetToken}`;
```
Abhijeet1005, line 150: Two things on these two lines:

**(1) Threat-model gap — the raw token still rides in the query string.** You hash the token at rest to survive a DB leak, but `?token=<raw>` here means the raw token lands in Express/proxy **access logs** and **Elastic APM** (`http.request.url`) in plaintext — the same "readable secret at rest" risk via a different sink. `Referrer-Policy: no-referrer` covers the Referer vector but not this one. Worth either scrubbing the `token` query param in the logging/APM layer or noting it as accepted in the PR description.

**(2) `baseUrl` can be `undefined`** (line 151): if neither `_system.site.url` nor `connectionInfo?.baseUrl` is set, this interpolates to `undefined/api/_internal/...` and emails a dead link. The old `resolveUrl(baseUrl!, …)` at least encoded that assumption with `!`. Consider throwing when `baseUrl` is missing rather than sending a broken link.

Abhijeet1005 (commented):
Thanks for tackling this — moving the reset token off client-rendered URLs is a solid, well-tested hardening change. I traced the full flow against `main` and the wiring holds up:

- Cookie path `/api/_internal/` is a prefix of the method endpoint `/api/_internal/method/...`, so the cookie is sent on the `resetPassword` call.
- `getCallContext` returns `HttpContext` and `cookieParser()` is registered, so `context.req.cookies` / `context.res.clearCookie` are populated at runtime.
- `registerModuleRoutes()` runs before `app.all('*')`, so the landing GET route is reachable.
- `Store.findOneAndDelete` returns the doc directly, and the TTL index (`expireAfterSeconds: 0`) cleans up tokens the landing route rejects without consuming.

**Particularly nice:**
- Atomic single-use via `findOneAndDelete` closes the find-then-delete TOCTOU window.
- The `createRouteHandler` reorder is a real latent-bug fix, not just a tidy-up — see the inline note.
- SHA-256-at-rest with a time-boxed plaintext fallback + a removal `TODO` is the right migration shape.

A few non-blocking notes inline (submitting as a comment). The main one I'd value your take on: the token still lives in the URL query string, given you went to the trouble of hashing it at rest.

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -116,19 +167,97 @@ export async function handleSendResetPasswordToken(args: Args, { connectionInfo
...
+      status: 302,
+      // Suppress the Referer so the token-bearing landing URL never leaks via the
+      // Referer header on any request that follows this navigation.
+      headers: { 'Referrer-Policy': 'no-referrer' },
+      redirect: resetPasswordUrl,
+    };
+  } catch (error) {
+    const message =
+      error instanceof Error
+        ? error.message
```
Abhijeet1005: When `params.query.token` is missing or non-string, `z.string().parse(...)` throws a `ZodError`, and `error instanceof Error` is `true` — so `message` becomes the ZodError's issues-array JSON, which gets URL-encoded into `?message=...` and shown to the user. The friendly `'This password reset link is invalid or has expired.'` fallback only fires for non-`Error` throws, which can't happen here. The "token query param is missing" test still passes because it only asserts `status=error`, not the message. Suggest throwing the friendly `Error` for the missing-token case (or not forwarding raw `error.message`), and asserting the user-facing message in that test.

packages/modelence/src/auth/resetPassword.ts
```diff
@@ -116,19 +167,97 @@ export async function handleSendResetPasswordToken(args: Args, { connectionInfo
...
+    const resetTokenDoc = await findResetTokenDoc(token);
+    if (!resetTokenDoc || resetTokenDoc.expiresAt < new Date()) {
+      throw new Error('This password reset link is invalid or has expired.');
+    }
+
+    // Hand the token to the SPA via an httpOnly cookie scoped to the API path,
+    // so it is only ever sent back on the resetPassword mutation request.
+    params.res.cookie(RESET_PASSWORD_COOKIE, token, {
+      httpOnly: true,
+      secure: process.env.NODE_ENV === 'production',
```
Abhijeet1005: Nit / consistency: this matches the existing `set-link-cookie` cookie, so no change required — just flagging that behind a TLS-terminating proxy where `NODE_ENV` isn't `production`, the cookie won't get the `Secure` attribute. If that's a supported deploy shape, deriving this from the resolved site URL's scheme (or a config flag) would be more robust than `NODE_ENV`.

packages/modelence/src/routes/handler.ts
```diff
@@ -51,17 +51,19 @@ export function createRouteHandler(method: string, path: string, handler: RouteH
       if (response) {
         res.status(response.status || 200);
 
-        if (response.redirect) {
-          res.redirect(response.redirect);
-        }
-
+        // Apply custom headers before sending the body/redirect, otherwise
```
Abhijeet1005, line 55: Nice fix, and worth calling out in the PR description: the previous order ran `res.redirect()` and **then** `res.send(response.data)`, so every redirect route (e.g. `verify-email`) was issuing a send-after-redirect (`ERR_HTTP_HEADERS_SENT`, swallowed by the surrounding try/catch) **and** dropping `response.headers`. This reorder fixes both the header drop and the double-send.

packages/modelence/src/auth/resetPassword.ts
omegascorp: This should be fine, the condition is for local development

packages/modelence/src/auth/resetPassword.ts
Abhijeet1005: Thanks for the fast turnaround. The telemetry redaction in `telemetry/index.ts` covers the APM / transaction-context vector well, and the fail-loud-on-missing-base-URL guard + the fixed friendly landing message both look spot on.

One residual specific to this thread, for completeness: the raw token is still in the URL query string (`?token=`), so a reverse proxy / web-server **access log** in front of the Node app (nginx, ALB, etc.) can still record it — that's upstream of anything app-level redaction can reach. It's arguably an acceptable tradeoff (it's an emailed GET link, and `verify-email` has the same shape), so this may be a "document it as accepted" rather than a code change — or a `token`-param scrub at the proxy log layer if you want belt-and-suspenders.

Also +1 on the `claimResetTokenById` commit-point change — deferring consumption until after validation/hashing, so a mid-flow failure doesn't burn the link, is a nicer design than consuming up front. 👍

omegascorp: @Abhijeet1005 thank you for the detailed comments; all should be addressed now. Regarding URLs with tokens in emails, I think it's an acceptable trade-off; most authentication libraries do it the same way.

Abhijeet1005:
Verified the latest commit (`6758cb1`) against the three new Bugbot findings — all look fully addressed:

- **Stored hash works as token (High):** removing the plaintext fallback so `findResetTokenDoc` only matches `hashToken(rawToken)` is the right fix. It also supersedes the "keep a short plaintext fallback for migration" approach discussed earlier in this thread (which I'd endorsed too) — that fallback was precisely what let a leaked SHA-256 digest be replayed as a bearer token, so dropping it and relying on the 1h TTL is the correct call. Only side effect I can see: a reset link emailed in the hour *before* hashing deploys would stop resolving until its TTL lapses — presumably an acceptable one-time rollout cost.
- **clearCookie omits matching flags (Medium):** sharing `RESET_PASSWORD_COOKIE_OPTIONS` across the set and clear calls is the correct way to guarantee the httpOnly cookie actually clears.
- **Cross-origin reset cookie (Medium):** `sameSite: 'lax'` plus `credentials: 'include'` on the method `fetch` covers both the email-link navigation and the XHR axes.

Thorough turnaround — looks good from my side.

artahian (approved):

PR #287, opened 2026-06-29
fix: restrict built-in file access to managed user paths

### LINKED ISSUE #286: Private file APIs are unauthenticated and can issue signed URLs or deletes for arbitrary paths
# Private file APIs are unauthenticated and can issue signed URLs or deletes for arbitrary paths

## Summary

The built-in `_system.files` methods appear to be exposed without any authentication or ownership checks, while the actual storage operations are proxied to Modelence Cloud using the application's server-side bearer token.

Because of that, any caller that can hit the method endpoint can request:

- upload URLs
- signed download URLs
- signed file URLs
- delete operations

for arbitrary `filePath` values, including `private/*` paths.

---

## Problem

In `packages/modelence/src/files/index.ts`, the `_system.files` module registers:

- `_system.files.getUploadUrl`
- `_system.files.getFileUrl`
- `_system.files.downloadFile`
- `_system.files.deleteFile`

These handlers call `callCloudApi(...)` directly and do not check `user`, `session`, or any file ownership rule before forwarding the request.

At the same time:

- `packages/modelence/src/methods/index.ts` treats methods with no declared permissions as callable without any additional access requirement
- `packages/modelence/src/app/backendApi.ts` sends the request to Modelence Cloud using `MODELENCE_SERVICE_TOKEN`, which is the app-level server credential rather than the caller's identity

So the file API is effectively acting as a public proxy to privileged storage operations.

---

## Why this is a problem

The documentation says private files are only accessible through time-limited signed URLs:

- `docs/web/files.mdx`

But right now, if a caller knows or can guess a `private/<path>` value, they may be able to:

- mint a signed URL for someone else's private file
- request a download URL for someone else's private file
- delete an arbitrary file path
- request upload credentials for arbitrary file paths

That means private file access is relying mostly on path secrecy, not on authentication or authorization.

---

## Suggested fix

At minimum, the built-in file methods should require an authenticated user for private file operations.

I think the safer direction would be:

1. Require authentication for `_system.files.*`
2. Deny anonymous access to `private/*`
3. Add an authorization/ownership policy for read, delete, and upload operations instead of forwarding arbitrary file paths directly
4. Add regression tests covering anonymous access and cross-user access for private files

---

## Scope

- `packages/modelence/src/files/index.ts`
- `packages/modelence/src/methods/index.ts`
- file-related auth/authorization tests

anirudh5harma:
Fixes modelence/modelence#286

## Summary

Built-in `_system.files` methods were forwarding storage operations with the app service token without checking who requested them. This scopes the browser-facing file APIs to authenticated users and managed per-user paths so private file access no longer depends on path secrecy.

Uploads through the built-in module now accept relative paths and are stored under `<visibility>/users/<user-id>/...`. Private file URL resolution and downloads require the authenticated owner, and deletes are limited to the caller's managed public and private paths.

The low-level server file helpers still exist for custom storage flows, but the docs now call out that callers must enforce their own auth and authorization when using those helpers directly.

## Tests

- `npm test -- --run src/files/index.test.ts`
- `npm test -- --run src/app/server.test.ts src/client/session.test.ts src/auth/session.test.ts`
- `npm run build`

---

> [!NOTE]
> **High Risk**
> Changes authorization for file storage APIs (prior path-based access could be abused); client apps may break if they used unprefixed stored paths or cross-user private paths.
> 
> **Overview**
> **Security fix for browser-facing `_system.files` operations:** built-in queries/mutations no longer forward arbitrary paths to Modelence Cloud with only the service token. They now require authentication where appropriate, normalize paths, and enforce per-user managed storage under `<visibility>/users/<user-id>/...`.
> 
> **Uploads** through the built-in mutation accept **relative** paths only (reject `public/`/`private/` prefixes and unsafe paths), require a signed-in user, and rewrite the cloud path to the caller’s managed namespace.
> 
> **Reads/deletes:** private `getFileUrl` and `downloadFile` require the authenticated owner; `downloadFile` rejects public paths. Public `getFileUrl` stays callable without a user. Deletes require auth and ownership for both public and private managed paths.
> 
> Low-level **`modelence/server`** helpers are unchanged; **docs** add a warning and updated client examples (managed paths, auth rules) and note that server callers must enforce their own authorization.
> 
> **Tests** cover auth failures, upload scoping, prefixed-path rejection, ownership checks, and public URL behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8111303cb07be74aca79a9fb46b98ac3aa74d90d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/web/files.mdx
+++ b/docs/web/files.mdx
@@ -10,6 +10,10 @@ Each file operation is available in two forms:
 - **Client (`modelence/client`)** — call directly from the browser via built-in methods
 - **Server (`modelence/server`)** — call from your server-side code (e.g. inside a mutation or cron job)
 
+<Warning>
+The browser-facing built-in methods are authenticated, managed file APIs. Uploads are automatically namespaced under `users/<user-id>/...`, and private file reads/deletes are restricted to the authenticated user's own managed paths.
+</Warning>
+
 ## Uploading a File
 
 <Tabs>
@@ -26,11 +30,13 @@ Each file operation is available in two forms:
       visibility: 'public',
     });
 
-    console.log(filePath); // e.g. 'public/photo.png'
+    console.log(filePath); // e.g. 'public/users/<user-id>/photo.png'
     ```
   </Tab>
   <Tab title="Server">
-    `getUploadUrl` returns a presigned URL and form fields. POST these together as `FormData` to upload the file directly to storage:
+    `getUploadUrl` returns a presigned URL and form fields. POST these together as `FormData` to upload the file directly to storage.
+
+    Since the server helper is low-level, apply your own auth/authorization rules before calling it:
 
     ```ts
     import { getUploadUrl } from 'modelence/server';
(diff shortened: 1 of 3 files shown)


omegascorp:
We were thinking about the solution to this, but chose another option. We've removed all frontend methods, so each application should authorize the action on the backend before performing it.

omegascorp:
Closing since the issue is solved in an alternative way

PR #290, opened 2026-07-07
MOD-453: feat: add magic link sign in

omegascorp:
* magic link sign-in and sign-up
* new method in store findOneAndUpsert

---

> [!NOTE]
> **High Risk**
> Large, security-critical authentication changes (token handling, cookies, sign-up, sessions, rate limits) plus a new DB uniqueness constraint that can fail startup on legacy duplicate emails.
> 
> **Overview**
> Adds **passwordless magic link authentication** (optional `auth.magicLink` + `email.magicLink`), including client APIs `sendMagicLink`, `loginWithMagicLink`, and `loginWithOneTimeCode`, server mutations, a scanner-safe landing route that stashes the token in an **httpOnly cookie**, and a `_modelenceMagicLinkTokens` store (hashed link + 6-digit code, single-use, rate limits). Optional **`allowSignup`** enables find-or-create sign-up with `provider: 'magicLink'` hooks; **`onBeforeSignup`** now covers `'magicLink'` as well as `'email'`.
> 
> Hardens related auth/data paths: **case-insensitive unique index** on `emails.address`, signup/magic-link handling of concurrent duplicate inserts, **`Store.findOneAndUpsert`** (`{ doc, isNew }`) used for atomic account creation, password reset as **set-or-reset** for OAuth/magic-link-only users (with magic link revocation on reset), and clearer logging when unique index builds fail on existing duplicates. Docs and nav updated; package version **0.21.0 → 0.22.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38a0f0b48a14a527c7c44c6a5f6b2cb235747394. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/modelence/src/app/authConfig.ts
```diff
@@ -257,6 +261,27 @@ export type AuthConfig = {
   /** @deprecated Use {@link AuthConfig.onAfterSignup} and {@link AuthConfig.onSignupError} instead. */
   signup?: AuthOption;
 
+  /**
+   * Enables passwordless magic link authentication. Disabled by default —
+   * magic link doubles as sign-up (an unknown email gets an account created
+   * when the link is used), so it must be opted into explicitly.
```
artahian: @omegascorp should we make this a boolean setting as well to explicitly opt-in for auto-creating accounts for unknown emails?

packages/modelence/src/auth/duplicateEmails.ts
```diff
@@ -0,0 +1,77 @@
...
+      // Deleted accounts are excluded: they do not participate in the partial
+      // unique index and must not be flagged for cleanup.
+      { $match: { 'emails.address': { $exists: true }, status: { $ne: 'deleted' } } },
+      { $unwind: '$emails' },
+      {
+        $group: {
+          _id: { $toLower: '$emails.address' },
+          userIds: { $addToSet: '$_id' },
+        },
+      },
```
artahian: Why do we need this? How can there be duplicate emails if there is an index?

artahian (changes requested):
The duplicate email logic seems strange to me - how can the unique index not exist? I don't think we should manually handle cases like this, the same can happen on every possible collection for every unique index. If we want to add handling for that, it should probably be a general logic for any unique index in that case.

packages/modelence/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "modelence",
-  "version": "0.21.0",
+  "version": "0.21.0-ml.dev.4",
```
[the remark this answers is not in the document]
omegascorp: Yes, it will be here until we merge it

omegascorp:
@artahian I've added a proper log for the unique index errors and removed the email deduplication logic; please check again

packages/modelence/src/auth/magicLink.ts
```diff
@@ -0,0 +1,642 @@
...
+  // that could otherwise skip this.
+  const emailProvider = getEmailConfig().provider;
+  if (!emailProvider) {
+    throw new Error('Email provider is not configured');
+  }
+
+  // Also a preflight: resolved before the token is stored, so a misconfigured
+  // server fails loudly without leaving behind an orphaned (unsendable) token.
+  // Without a base URL we'd email a broken `undefined/...` link.
+  const baseUrl = (getConfig('_system.site.url') as string | undefined) || connectionInfo?.baseUrl;
```
modelence-ops: This fallback reintroduces the Host-header issue we fixed in the signup flow (`MODELENCE_SITE_URL` vs raw Host): `connectionInfo.baseUrl` is derived from the request, so a spoofed Host header controls the URL we email — and here the link *is* the credential (reset-poisoning applied to login). The magic-link docs in this PR already say we hard-fail without `_system.site.url`; let's make the code match and drop the fallback (and the fallback test). Same applies to the `req.get('host')` fallback in `handleMagicLinkLanding` below.

packages/modelence/src/auth/magicLink.ts
```diff
@@ -0,0 +1,642 @@
...
+
+  // Clicking the emailed link proves ownership of the address. A concurrent
+  // password signup may have created the account as unverified — this link is
+  // the proof that verifies it.
+  const emailDoc = userDoc.emails?.find((e) => e.address.toLowerCase() === tokenDoc.email);
+  const wasUnverified = !emailDoc?.verified;
+  await usersCollection.updateOne(
+    { _id: userDoc._id, 'emails.address': tokenDoc.email },
+    { $set: { 'emails.$.verified': true } }
+  );
```
modelence-ops: This `updateOne` has no collation, but the lookup that found the user uses strength-2. `tokenDoc.email` is lowercased, and OAuth-created accounts can store the address with its original casing — so the positional update silently matches nothing while `wasUnverified` (which does handle casing via `toLowerCase()` above) still fires `onAfterEmailVerification`. Add the same collation to the write.

Related: the hooks below receive the pre-update `userDoc`, so they see `verified: false` even after a successful verification.

packages/modelence/src/auth/magicLink.ts
```diff
@@ -0,0 +1,642 @@
...
+  if (wasUnverified) {
+    authConfig.onAfterEmailVerification?.({
+      provider: 'magicLink',
+      user: userDoc,
+      session,
+      connectionInfo,
+    });
+  }
+
+  authConfig.onAfterLogin?.({
```
modelence-ops: `onAfterLogin` and `onAfterEmailVerification` are fired without `await`, while `onBeforeSignup` is awaited. If a user's async hook rejects, that's an unhandled promise rejection (can crash the process depending on Node config). Either await them or explicitly `.catch` + log. Worth checking whether the password-reset path has the same pattern.

packages/modelence/src/auth/db.ts
```diff
@@ -126,3 +137,32 @@ export const resetPasswordTokensCollection = new Store('_modelenceResetPasswordT
...
+  },
+  indexes: [
+    {
+      key: { token: 1 },
+      unique: true,
+    },
+    {
+      key: { expiresAt: 1 },
+      expireAfterSeconds: 0,
+    },
```
modelence-ops, line 173: +1 to the Bugbot finding here: `handleLoginWithOneTimeCode` looks up by `{ email, code }` and every wrong guess runs `updateMany({ email })` — both collection scans without an `{ email: 1 }` index. Rate-limited, but it's still an attacker-controllable hot path and the index is cheap.

packages/modelence/src/auth/magicLink.ts
```diff
@@ -0,0 +1,642 @@
...
+  const magicLinkUrl = `${baseUrl}/api/_internal/auth/magic-link?token=${magicLinkToken}`;
+
+  // Send email
+  const template = getEmailConfig()?.magicLink?.template || magicLinkTemplate;
+  const htmlTemplate = template({ email, magicLinkUrl, code: oneTimeCode, name: '' });
+  const textContent = htmlToText(htmlTemplate);
+
+  await emailProvider.sendEmail({
+    to: email,
+    from: getEmailConfig()?.from || 'noreply@modelence.com',
```
modelence-ops: Let's throw here instead of falling back to `noreply@modelence.com` — customer apps would send their auth email from our domain (and fail SPF/DKIM anyway). Same reasoning as the missing-provider check above.

modelence-ops (commented):
Non-blocking comments — the two I'd prioritize are the base-URL fallback (Host-header poisoning of the emailed link) and the missing collation on the verify update.

packages/modelence/src/auth/magicLink.ts
```diff
@@ -0,0 +1,642 @@
...
+        error,
+        session,
+        connectionInfo,
+      });
+    }
+    throw error;
+  }
+}
+
+async function signupNewUser(params: MagicLinkAuthParams) {
```
modelence-ops: Bigger-picture suggestion: since the magic link proves email ownership, this whole branch is a find-or-create — which Mongo can do in one atomic op:

```ts
const result = await usersCollection.findOneAndUpdate(
  { emails: { $elemMatch: { address: email } } },
  {
    $setOnInsert: {
      handle: resolvedHandle,
      status: 'active',
      emails: [{ address: email, verified: true }],
      createdAt: new Date(),
      authMethods: {},
    },
  },
  {
    upsert: true,
    returnDocument: 'after',
    includeResultMetadata: true,
    collation: { locale: 'en', strength: 2 },
  }
);
const isNewUser = Boolean(result.lastErrorObject?.upserted);
```

`isNewUser` picks between the signup and login tails, and the claim→insert→catch-duplicate→recovery-login→restore-token chain collapses: no `restoreClaimedMagicLinkToken`, no duplicate-key recovery branch, no wrong-hook-on-recovery edge cases. Probably deletes a few hundred lines here plus all the tests that only exist to exercise the compensation paths.

Two implementation notes: `$elemMatch` (rather than `'emails.address'`) avoids the upsert field-inference conflict with `$setOnInsert: { emails: ... }`, and concurrent upserts can still race to insert (known Mongo caveat) — a single retry on E11000 covers it, which is still far simpler than the current saga. Token claim stays where it is (before this op), and handle resolution moves just above it, same as now.

artahian (approved):

PR #299, opened 2026-07-16
chore: remove unused React imports

aadityakumarsah:
Resolves #298

Removed unused `import React from 'react'` in `AppProvider.tsx`, `AppProvider.test.tsx`, and `queryProvider.tsx` to optimize module resolution.

---

> [!NOTE]
> **Medium Risk**
> Auth middleware behavior change affects who can open WebSocket connections; React import edits are low risk.
> 
> **Overview**
> Removes default **`React`** imports from **`AppProvider`**, **`queryProvider`**, and the AppProvider test in favor of named hooks/types only (modern JSX transform).
> 
> **Socket.IO connection middleware** no longer calls **`next()`** in a **`finally`** block after **`authenticate`** fails. On success it sets **`socket.data`** and calls **`next()`**; on failure it calls **`next(error)`** so invalid tokens reject the handshake instead of connecting with unset data. A unit test covers the rejection path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b074857e00a108427f5f05cfad1a6c972ad5f747. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #302, opened 2026-07-17
fix: cron unhandled promises

### LINKED ISSUE #303: Fix unhandled promise rejections and optimize array iterations in cron jobs
The `tickCronJobs` and `startCronJobs` functions in `packages/modelence/src/cron/jobs.ts` currently use standard `Array.prototype.forEach` to loop through cron job records and objects.

In `tickCronJobs`, this is problematic because it uses an `async` callback inside a synchronous `forEach` loop. The `forEach` loop does not wait for the promises to resolve, meaning the asynchronous execution of `runCronJob` runs detached from the tick cycle. This can potentially lead to unhandled promise rejections crashing the Node.js process and creates a race condition if multiple intervals tick before the previous jobs finish safely.

Additionally, using `forEach` across the file introduces minor but unnecessary closure allocation overhead in a fast-running interval.

**Proposed Solution:**
- Refactor the `forEach(async (job) => { ... })` loop in `tickCronJobs` to use `await Promise.all(Object.values(cronJobs).map(async (job) => { ... }))`. This allows concurrent execution of the jobs while properly awaiting all of them before the tick cycle completes.
- Replace synchronous `forEach` loops over arrays and object values in `startCronJobs` with optimized `for...of` loops to reduce closure allocations and improve readability.

aadityakumarsah:
Resolves #303

This PR addresses two interconnected issues within the `packages/modelence/src/cron/jobs.ts` file:
1. **Unhandled Promises / Race Conditions:** Replaced the `forEach(async ...)` loop inside `tickCronJobs` with an `await Promise.all(...)` wrapper mapped over the object values. This ensures that the interval correctly awaits the completion of all concurrently scheduled cron jobs for the current tick, properly trapping any rejections that might otherwise crash the process and preventing runaway job accumulations.
2. **Loop Optimization:** Replaced synchronous `.forEach` calls in `startCronJobs` with standard `for...of` loops, avoiding unnecessary callback closure allocations.

**Testing:**
- Existing unit tests for cron jobs and timing pass successfully.
- Linters run successfully without errors.


omegascorp:
@aadityakumarsah this seems to contain changes from another PR, though it will likely be resolved automatically when the other PR is merged

omegascorp:
Thanks for this, and sorry for the slow response. Closing — the socket.io fix and React import removals landed in #299, and the branch is conflicting now. On the cron change, see #303: that rejection path is already guarded.

Your instinct pointed at something real one line up though — opened #338 for it. A small PR against fresh main touching only jobs.ts would be welcome.

The dependency cleanup looks fine as its own PR, minus the root package-lock.json (looks accidental).

PR #305, opened 2026-07-17
fix: mitigate password timing attack and email enumeration in login

### LINKED ISSUE #304: Mitigate password timing attack and email enumeration in login
In `packages/modelence/src/auth/login.ts` inside the `handleLoginWithPassword` function, an early return/throw occurs if the user is not found or does not have a password hash:

```typescript
const passwordHash = userDoc?.authMethods?.password?.hash;
if (!passwordHash) {
  throw incorrectCredentialsError();
}
```

If the user does exist and has a password, the code proceeds to evaluate `bcrypt.compare`:

```typescript
const isValidPassword = await bcrypt.compare(password, passwordHash);
```

Because `bcrypt.compare` is designed to be computationally expensive (taking tens to hundreds of milliseconds depending on the cost factor), an attacker can measure the time it takes for the server to respond to a login request. 
- Fast response = user does not exist (or has no password).
- Slow response = user exists.

This allows an attacker to enumerate which email addresses are registered in the system.

Additionally, the code previously checked if the email address was verified *before* checking if the password was correct. This allowed an attacker to determine if an email was registered but unverified simply by providing any dummy password.

**Proposed Solution:**
1. Always perform a dummy `bcrypt.compare` against a constant hash if the user or `passwordHash` is not found. This ensures the function takes approximately the same amount of time regardless of whether the email exists.
2. Move the email verification check to occur *after* the password has been successfully verified, so verification status is only revealed to users who know the correct password.

aadityakumarsah:
Resolves #304

## Summary

- Runs `bcrypt.compare` against a generated dummy hash (no hardcoded value) when no password hash is found, keeping unknown-email and known-email login timing comparable.
- Checks email verification only after the password succeeds, preventing unverified-account enumeration.

## Behavioral change

Unverified users who type a **wrong password** now get `Incorrect email/password combination` instead of the "email isn't verified" message. That is intentional (prevents enumeration), but it removes a self-explanatory hint for legitimate unverified users and may generate support tickets. Worth calling out in release notes.

## Validation

- `npm run lint:check`
- `npm run build`
- `npx tsc --noEmit`
- `npm test` (859 passing)

### DIFF
--- a/packages/modelence/src/auth/login.ts
+++ b/packages/modelence/src/auth/login.ts
@@ -35,6 +35,8 @@ function isEmailVerificationRequired(): boolean {
   return Boolean(getConfig('_system.user.auth.email.verification'));
 }
 
+const DUMMY_PASSWORD_HASH = '$2b$10$AUM.KxS9LpsdJ5XQ85jQG.zC2y8b375b4pQeO2XoGf29xNn4/WlFm';
+
 export async function handleLoginWithPassword(
   args: Args,
   { user, session, connectionInfo, res }: Context
@@ -69,10 +71,17 @@ export async function handleLoginWithPassword(
     );
 
     const passwordHash = userDoc?.authMethods?.password?.hash;
-    if (!passwordHash) {
+
+    // Always run bcrypt.compare to mitigate timing attacks that could reveal if an email exists.
+    const hashToCompare = passwordHash || DUMMY_PASSWORD_HASH;
+    const isValidPassword = await bcrypt.compare(password, hashToCompare);
+
+    if (!passwordHash || !isValidPassword) {
       throw incorrectCredentialsError();
     }
 
+    // Only check email verification if the user exists and the password is correct,
+    // to prevent email enumeration by observing verification errors before password checks.
     const emailDoc = userDoc.emails?.find((e) => e.address.toLowerCase() === email);
 
     if (!emailDoc?.verified && isEmailVerificationRequired()) {
@@ -80,11 +89,6 @@ export async function handleLoginWithPassword(
         "Your email address hasn't been verified yet. Please check your inbox for the verification email.",
         'EMAIL_NOT_VERIFIED'
       );
-    }
-
-    const isValidPassword = await bcrypt.compare(password, passwordHash);
-    if (!isValidPassword) {
-      throw incorrectCredentialsError();
     }
 
     await setSessionUser(session.authToken, userDoc._id);
--- a/packages/modelence/src/auth/login.ts
+++ b/packages/modelence/src/auth/login.ts
@@ -1,3 +1,5 @@
+import { randomBytes } from 'crypto';
+
 import bcrypt from 'bcrypt';
 import { z } from 'zod';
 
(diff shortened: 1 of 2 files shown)


aadityakumarsah:
The build was failing because of a TypeScript error in `src/auth/session.ts` that was carried over from the `fix/structured-logging` branch (it inferred `hashedToken` could be null inside the session update block). I just pushed a commit to fix the types, so the build and tests should pass now!

AtulYadav25:
Good to see your contributions, @aadityakumarsah ! 👏

One small suggestion: try creating a new branch for each PR instead of reusing the same on

rigth now, commits from previous PRs are showing up in this PR history too, this will help reviewers go through it quicker

omegascorp:
I agree with @AtulYadav25, @aadityakumarsah. Please update the PR to contain only relevant changes, and please also resolve the conflicts.

packages/modelence/src/auth/login.ts
```diff
@@ -69,10 +71,17 @@ export async function handleLoginWithPassword(
     );
 
     const passwordHash = userDoc?.authMethods?.password?.hash;
-    if (!passwordHash) {
+
+    // Always run bcrypt.compare to mitigate timing attacks that could reveal if an email exists.
+    const hashToCompare = passwordHash || DUMMY_PASSWORD_HASH;
```
omegascorp, line 79: let's generate a hash instead of hardcoding it
```
bcrypt.hashSync(randomBytes(32).toString('hex'), SALT_ROUNDS); // SALT_ROUNDS = 10

```

omegascorp (commented):
*Behavioral regression for legitimate unverified users.* Previously an unverified user typing the wrong password got "your email isn't verified." Now they get "incorrect email/password combination" and have no path to discover the actual problem. That's correct security-wise but it's a support-ticket generator. It's the accepted cost of the fix — but it should be called out in the changelog, since it changes a user-facing error for a non-attacker case.

*The enumeration hole this closes is still wide open elsewhere.* signup.ts:56 throws User with email already exists: ${existingEmail?.address} — a direct, unauthenticated, non-timing-based confirmation that an email is registered. Same at line 138. So an attacker who wants to enumerate accounts will just hit signup and get a definitive answer instantly, rather than doing statistical timing analysis on login. Note that resetPassword.ts:99-101 gets this right and returns a generic success. This doesn't make PR #305 wrong, but it means the timing half delivers little practical benefit until signup is fixed too. Worth a follow-up issue.

*Test coverage has a gap.* The added tests are good and target the right behaviors. Missing: the case where userDoc exists but has no authMethods.password at all (OAuth-only or magic-link-only account). That path now goes through the dummy compare, and it's exactly where the redundant !passwordHash guard earns its keep — worth pinning with a test.

packages/modelence/src/auth/login.ts
aadityakumarsah: Done in 116c497f: replaced the hardcoded hash with `bcrypt.hashSync(randomBytes(32).toString("hex"), SALT_ROUNDS)` (SALT_ROUNDS = 10). I also addressed the rest of your review:

- Added a test pinning the dummy-hash compare for accounts with no `authMethods.password` (OAuth-only / magic-link-only), 8837f3e1.
- Called out the user-facing error change for unverified users in the PR description (this repo has no CHANGELOG file; releases use GitHub Releases).
- Created follow-up issue #339 for the signup.ts:56/138 email enumeration hole.

omegascorp (approved):
looks good, moving forward

PR #310, opened 2026-07-18
fix(auth): prevent login/signup when already authenticated

### LINKED ISSUE #311: Bug: login and signup handlers do not reject already authenticated users
Currently, the `login` and `signup` handlers do not check if a user is already authenticated. This allows active sessions to hit these endpoints, which can cause unexpected state overlaps, double rate-limit consumption, and potential security issues.

These handlers should verify the session context and throw an error early if a user is already logged in.

aadityakumarsah:
## Description
This PR implements missing checks in the `login` and `signup` handlers to properly reject requests from users who are already actively authenticated. 

**Changes:**
- Added a check for `user` in `src/auth/login.ts` and `src/auth/signup.ts`.
- Throws an `Error` if the user context already exists, terminating the authentication attempt safely.

This ensures proper state management and mitigates unintended side-effects from duplicate login/signup attempts by active sessions.

## Testing
- Tests in `packages/modelence` passed successfully (`vitest run`).

Fixes #311

### DIFF
--- a/packages/modelence/src/auth/login.ts
+++ b/packages/modelence/src/auth/login.ts
@@ -60,7 +60,7 @@ export async function handleLoginWithPassword(
     // TODO: add rate limiting by email (and perhaps IP address overall)
 
     if (user) {
-      // TODO: handle cases where a user is already logged in
+      throw new Error('You are already logged in.');
     }
 
     const userDoc = await usersCollection.findOne(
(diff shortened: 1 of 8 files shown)


omegascorp:
@aadityakumarsah this has the same issue as the other PRs, it contains fixes from other ones, so I can't review it

PR #313, opened 2026-07-18
fix(auth): rate limit profile updates

aadityakumarsah:
## Summary

- add per-user limits for profile updates (30 per 15 minutes; 200 per day)
- expose `auth.rateLimits.updateProfile` for configuration overrides
- validate profile updates and handle availability before consuming the quota
- add coverage for defaults, overrides, validation failures, and successful updates

This branch was rebuilt from current `main` and contains only the profile-update rate-limit change.

## Testing

- `npm run lint:check`
- `npm run build`
- `npm test`

---

> [!NOTE]
> **Low Risk**
> Scoped to profile update mutation with conservative defaults; quota is not consumed on validation failures, limiting user impact.
> 
> **Overview**
> Adds **per-user rate limiting** on profile updates so abusive or automated `updateProfile` traffic is capped like other auth actions.
> 
> Default limits are **30 updates per 15 minutes** and **200 per day** per user, registered on the auth module’s rate-limit rules. Apps can tune them via **`auth.rateLimits.updateProfile`** using the same override/merge behavior as other auth buckets.
> 
> `handleUpdateProfile` now calls **`consumeRateLimit`** with bucket `updateProfile` only when there is a non-empty update to persist—**after** field validation, the optional `validateProfileUpdate` hook, and handle uniqueness checks—so invalid or rejected requests do not burn quota. Tests cover defaults, overrides, and when the limit is (not) consumed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1af3bfb689181078a5fdbfcd410e118c6f985c50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/profile.ts
+++ b/packages/modelence/src/auth/profile.ts
@@ -3,6 +3,7 @@ import { Args, Context, UpdateProfileProps } from '../methods/types';
 import { serializeUserForClient } from './utils';
 import { validateProfileFields } from './validators';
 import { getAuthConfig } from '@/app/authConfig';
+import { consumeRateLimit } from '../rate-limit/rules';
 
 export async function getOwnProfile(props: Args, { user }: Context) {
   if (!user) {
@@ -49,6 +50,8 @@ export async function handleUpdateProfile(props: Args, { user }: Context) {
 
   //Update profile
   if (Object.keys(update).length > 0) {
+    await consumeRateLimit({ bucket: 'updateProfile', type: 'user', value: user.id });
+
     const setFields: Record<string, unknown> = {}; //Used to set the fields
     const unsetFields: Record<string, ''> = {}; //Used to clear the fields the mongodb, because mongodb rejects undefined values when updating fields
 
--- a/packages/modelence/src/auth/user.ts
+++ b/packages/modelence/src/auth/user.ts
@@ -55,6 +55,8 @@ function defaultAuthRateLimits(): RateLimitRule[] {
     { bucket: 'oneTimeCode', type: 'ip', window: time.days(1), limit: 100 },
     { bucket: 'oneTimeCode', type: 'email', window: time.hours(1), limit: 10 },
     { bucket: 'oneTimeCode', type: 'email', window: time.days(1), limit: 20 },
+    { bucket: 'updateProfile', type: 'user', window: time.minutes(15), limit: 30 },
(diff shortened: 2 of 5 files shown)


packages/modelence/src/auth/profile.ts
```diff
@@ -21,11 +22,13 @@ export async function getOwnProfile(props: Args, { user }: Context) {
   };
 }
 
-export async function handleUpdateProfile(props: Args, { user }: Context) {
+export async function handleUpdateProfile(props: Args, { user, connectionInfo }: Context) {
```
AtulYadav25: Unused `connectionInfo`

packages/modelence/src/auth/session.ts
```diff
@@ -141,10 +139,16 @@ export function clearAuthTokenCookie(res: Response) {
   });
 }
 
+import { consumeRateLimit } from '../rate-limit/rules';
```
AtulYadav25: Please move the import statement to the top, follow clean code

packages/modelence/src/auth/user.ts
```diff
@@ -107,8 +113,13 @@ export function buildAuthRateLimits(config: AuthRateLimitsConfig = {}): RateLimi
   return merged;
 }
 
-export async function createGuestUser() {
-  // TODO: add rate-limiting and captcha handling
+import { Context } from '../methods/types';
```
AtulYadav25: Please move import statement to the top

omegascorp:
@aadityakumarsah why do we need a rate limit on `sessionInit`? This basically means that when a lot of users visit from the same IP, it will start blocking them. Example of a false positive failure: a large company is using a product built on Modelence, and hundreds of users are visiting from the same corporate IP

aadityakumarsah:
@omegascorp You are right. `sessionInit` is the wrong place for an IP-based rate limit: clients behind corporate/carrier NATs could be blocked, and the session is already created before the handler runs, so it does not reliably cap session creation. I removed the `sessionInit` limiter and its configuration while resolving the branch conflicts. The remaining profile-update limit is keyed by authenticated user ID, not IP.

omegascorp:
@aadityakumarsah this branch seems to have the same issue as the other branches - it has unrelated changes

omegascorp (approved):

PR #314, opened 2026-07-18
fix: remove empty unload listener to support BFCache

### LINKED ISSUE #308: Remove unload listener to prevent BFCache performance degradation
The `packages/modelence/src/client/renderApp.tsx` file contains an empty `window.addEventListener('unload', () => {});` with a comment stating it "prevents bfcache in most browsers."

While this forces a fresh render on back-navigation, disabling the Back-Forward Cache (bfcache) severely degrades the user experience by artificially inflating load times when users navigate back or forward using their browser controls.

---

### ⚠️ Removing the listener alone is not sufficient

The listener was **added deliberately** in `70a40aba` ("Fix bfcache issues", Dec 2024) — the same commit that set `Cache-Control: no-store` in `vite.ts`. Disabling bfcache *was* the intended behavior, because the client has no way to revalidate a restored session.

Deleting the two lines without a restore-time revalidation reintroduces an **auth regression**:

1. User logs out — `logout()` clears the server session, the auth token, and the Zustand `user` store
2. User presses Back — bfcache restores the frozen JS heap, `user` included
3. `renderApp` does **not** re-run on restore, so nothing re-evaluates the session
4. The app renders as **logged in**, and every subsequent request fails against the now-invalid token

All of the state cleared by logout is in-memory or `localStorage`; none of it survives a bfcache restore as "cleared". There is currently **no `pageshow` / `persisted` handling anywhere in the codebase**.

### Proposed Solution

Remove the `unload` listener **and** add restore-time session revalidation, in a single change:

- [ ] Remove `window.addEventListener('unload', () => {});` from `renderApp.tsx`
- [ ] Add a `pageshow` listener in `renderApp`, gated on `event.persisted`
- [ ] On that restore path, re-evaluate the session (refresh config, auth token, and current user) so a page restored after logout becomes anonymous before any further requests
- [ ] Add a regression test for the logout → Back → restored-page path asserting the user is anonymous
- [ ] Re-evaluate whether the `Cache-Control: no-store` headers added alongside the listener in `vite.ts` are still needed

**Note:** `pagehide` is the correct hook for teardown on the way out, but the restore side must use `pageshow` with `event.persisted` — that is where the stale session has to be corrected.

### Prior attempts

Three PRs have attempted the two-line deletion without the revalidation half, and each would ship the regression above:

- #309 (closed)
- #314 (open)
- `ba65f7fe` (never merged)

In [this comment](https://github.com/modelence/modelence/issues/308#issuecomment-5045399488), @aadityakumarsah indicated the `pageshow` revalidation was already implemented — that work should be folded into whichever PR lands, rather than opening a fourth deletion-only PR.

ankit-0903:
### What & Why
This PR removes the empty `unload` event listener and its comment from `renderApp.tsx`. 
The presence of an `unload` listener, even when empty, prevents browsers from caching the page in the Back-Forward Cache (bfcache), resulting in slower navigation times when users use back/forward controls. 

Removing this listener enables bfcache optimization in client applications and resolves issue #308.

Closes #308

### How it was tested
1. Verified that the package compiles and builds successfully via `npm run build`.
2. Verified that all unit tests in `src/client/renderApp.test.tsx` pass successfully.
3. Verified the entire test suite (777 tests) passes successfully.

---

> [!NOTE]
> **Medium Risk**
> Touches session initialization and auth token handling on navigation restore; behavior is covered by new unit tests but affects every client app using `renderApp`.
> 
> **Overview**
> Removes the empty **`unload`** listener in **`renderApp`** so browsers can keep pages in the back-forward cache, and replaces it with a **`pageshow`** handler that runs only when **`event.persisted`** is true.
> 
> On BFCache restore, **`revalidateSessionOnPageShow`** calls **`_system.session.init`** again (via shared **`applySessionPayload`**), refreshes auth/config/user state (including clearing the user after logout), restarts the session heartbeat, and triggers **`revalidateModelenceQueries`** to invalidate the connected TanStack Query cache so UI data is not stale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42822667c0b77140605d674b32801084cfdc677e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/client/session.ts
+++ b/packages/modelence/src/client/session.ts
@@ -145,14 +145,7 @@ export function _isReconciliationPending(): boolean {
   return pendingReconciliation;
 }
 
-export async function initSession() {
-  if (isInitialized) {
-    return;
-  }
-
-  isInitialized = true;
-
-  const payload = await callMethod<SessionInitPayload>('_system.session.init');
+function applySessionPayload(payload: SessionInitPayload) {
   const { configs, session, user } = payload;
   setupRequired = Boolean(payload.setupRequired);
   _setConfig(configs);
@@ -165,29 +158,36 @@ export async function initSession() {
   }
 
   useSessionStore.getState().setUser(parseUser(user));
+}
+
+export async function initSession() {
+  if (isInitialized) {
+    return;
+  }
+
+  isInitialized = true;
+
+  const payload = await callMethod<SessionInitPayload>('_system.session.init');
+  applySessionPayload(payload);
 
   await loopSessionHeartbeat();
 }
 
 /** Revalidate session on pageshow when restored from BFCache (`event.persisted === true`). */
 export async function revalidateSessionOnPageShow() {
+  if (heartbeatTimer !== null) {
+    clearTimeout(heartbeatTimer);
+    heartbeatTimer = null;
+  }
+
   try {
     const payload = await callMethod<SessionInitPayload>('_system.session.init');
-    const { configs, session, user } = payload;
-    setupRequired = Boolean(payload.setupRequired);
-    _setConfig(configs);
-
-    const config = getClientConfig();
-    if (config) {
-      config.setAuthToken(session.authToken);
-    } else {
-      setLocalStorageSession(session);
-    }
-
-    useSessionStore.getState().setUser(parseUser(user));
+    applySessionPayload(payload);
     revalidateModelenceQueries();
   } catch (error) {
     console.error('Modelence: session revalidation failed on pageshow', error);
+  } finally {
+    void loopSessionHeartbeat();
   }
 }
 
--- a/packages/modelence/src/client/session.ts
+++ b/packages/modelence/src/client/session.ts
@@ -4,6 +4,7 @@ import { callMethod } from './method';
 import { _setConfig } from '../config/client';
 import { getLocalStorageSession, setLocalStorageSession } from './localStorage';
 import { getClientConfig } from './clientConfig';
+import { revalidateModelenceQueries } from './query';
 import { time } from '../time';
 import { Configs } from '../config/types';
 
@@ -166,6 +167,28 @@ export async function initSession() {
   useSessionStore.getState().setUser(parseUser(user));
 
   await loopSessionHeartbeat();
+}
+
+/** Revalidate session on pageshow when restored from BFCache (`event.persisted === true`). */
+export async function revalidateSessionOnPageShow() {
+  try {
+    const payload = await callMethod<SessionInitPayload>('_system.session.init');
+    const { configs, session, user } = payload;
+    setupRequired = Boolean(payload.setupRequired);
+    _setConfig(configs);
+
+    const config = getClientConfig();
+    if (config) {
+      config.setAuthToken(session.authToken);
+    } else {
+      setLocalStorageSession(session);
+    }
+
+    useSessionStore.getState().setUser(parseUser(user));
+    revalidateModelenceQueries();
+  } catch (error) {
+    console.error('Modelence: session revalidation failed on pageshow', error);
+  }
 }
 
 /** Idempotent, client-only. Auto-started by `initSession`; call explicitly after `hydrateSession`. */
--- a/packages/modelence/src/client/query.test.ts
+++ b/packages/modelence/src/client/query.test.ts
@@ -11,7 +11,12 @@ vi.doMock('../websocket/client', () => ({
   subscribeLiveQuery: mockSubscribeLiveQuery,
 }));
 
-const { modelenceLiveQuery, disconnectModelenceQueryClient } = await import('./query');
+const {
+  modelenceLiveQuery,
+  disconnectModelenceQueryClient,
+  connectModelenceQueryClient,
+  revalidateModelenceQueries,
+} = await import('./query');
 
 describe('modelenceLiveQuery', () => {
   const originalWindow = globalThis.window;
@@ -23,6 +28,19 @@ describe('modelenceLiveQuery', () => {
 
   afterEach(() => {
     globalThis.window = originalWindow;
+  });
+
+  test('revalidateModelenceQueries invalidates queries on connected QueryClient', () => {
+    const mockInvalidateQueries = vi.fn();
+    const fakeQueryClient = {
+      getQueryCache: () => ({ subscribe: vi.fn() }),
+      invalidateQueries: mockInvalidateQueries,
+    } as unknown as Parameters<typeof connectModelenceQueryClient>[0];
+
+    connectModelenceQueryClient(fakeQueryClient);
+    revalidateModelenceQueries();
+
+    expect(mockInvalidateQueries).toHaveBeenCalledTimes(1);
   });
 
   test('queryFn returns a never-resolving promise on the server (no window)', async () => {
--- a/packages/modelence/src/client/query.ts
+++ b/packages/modelence/src/client/query.ts
@@ -76,6 +76,13 @@ export function disconnectModelenceQueryClient() {
   websocketsStarted = false;
 }
 
+/** Revalidate all cached queries when restored from BFCache (`event.persisted === true`). */
+export function revalidateModelenceQueries() {
+  if (queryClientRef) {
+    void queryClientRef.invalidateQueries();
+  }
+}
+
 /** @deprecated Use `connectModelenceQueryClient(queryClient)` instead. */
 export class ModelenceQueryClient {
   connect(queryClient: QueryClient) {
--- a/packages/modelence/src/client/renderApp.test.tsx
+++ b/packages/modelence/src/client/renderApp.test.tsx
@@ -109,7 +109,7 @@ describe('client/renderApp', () => {
     globalThis.window = originalWindow;
   });
 
-  test('initializes error handler, attaches unload listener, and renders AppProvider', () => {
+  test('initializes error handler and renders AppProvider', () => {
     renderApp({
       loadingElement: <div>Loading</div>,
       routesElement: <div>Routes</div>,
@@ -122,7 +122,6 @@ describe('client/renderApp', () => {
     const renderFn = rootResult ? (rootResult.value as { render: Mock }).render : undefined;
     expect(renderFn).toBeDefined();
     expect(renderFn).toHaveBeenCalledTimes(1);
-    expect(addEventListenerMock).toHaveBeenCalledWith('unload', expect.any(Function));
   });
 
   test('creates favicon link when none exists', () => {
--- a/packages/modelence/src/client/renderApp.test.tsx
+++ b/packages/modelence/src/client/renderApp.test.tsx
@@ -25,10 +25,12 @@ vi.doMock('react-dom/client', () => ({
 
 const mockHydrateSession = vi.fn();
 const mockStartSessionHeartbeat = vi.fn();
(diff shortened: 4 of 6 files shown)


omegascorp (changes requested):
Blocking this change for now, please check the updated description of the task

ankit-0903:
@omegascorp I've updated the PR as requested in the updated issue
description!

   - Removed the empty unload listener to enable BFCache across browsers.
   - Added a pageshow event listener (event.persisted === true) that calls
   revalidateSessionOnPageShow() to re-fetch _system.session.init, updating
   session config, auth token, and current user. Added
   revalidateModelenceQueries() to call queryClientRef.invalidateQueries() on
   BFCache restore, ensuring stale cached queries from previous sessions are
   immediately invalidated/refetched.
   - Added unit tests for pageshow handling and query revalidation. All CI
   checks are passing!

packages/modelence/src/client/session.ts
```diff
@@ -168,6 +169,28 @@ export async function initSession() {
...
+    if (config) {
+      config.setAuthToken(session.authToken);
+    } else {
+      setLocalStorageSession(session);
+    }
+
+    useSessionStore.getState().setUser(parseUser(user));
+    revalidateModelenceQueries();
+  } catch (error) {
+    console.error('Modelence: session revalidation failed on pageshow', error);
```
omegascorp, line 188: **The heartbeat loop is never re-armed on restore.**

`loopSessionHeartbeat` chains itself with `setTimeout` and `heartbeatTimer` is a module-level non-null handle. Timers are suspended while the page is frozen in bfcache. If the page sat frozen past the server's session TTL, the chain can be effectively dead while `heartbeatTimer !== null` still guards `startSessionHeartbeat` from ever restarting it — so the session silently stops being kept alive after a back-navigation restore.

Since this function already re-runs `_system.session.init`, it should re-arm the heartbeat the way `initSession` does:

```suggestion
    useSessionStore.getState().setUser(parseUser(user));
    revalidateModelenceQueries();

    if (heartbeatTimer !== null) {
      clearTimeout(heartbeatTimer);
      heartbeatTimer = null;
    }
    void loopSessionHeartbeat();
```

packages/modelence/src/client/session.ts
```diff
@@ -168,6 +169,28 @@ export async function initSession() {
...
+    setupRequired = Boolean(payload.setupRequired);
+    _setConfig(configs);
+
+    const config = getClientConfig();
+    if (config) {
+      config.setAuthToken(session.authToken);
+    } else {
+      setLocalStorageSession(session);
+    }
+
```
omegascorp: **This duplicates `initSession` verbatim.**

Lines 174-186 are a copy of `initSession`'s body (lines 155-167): same `callMethod`, same `setupRequired`, same `_setConfig`, same `getClientConfig()`/`setLocalStorageSession` branch, same `setUser`. Counting `reconcileSession`, there are now three copies of this sequence in the file.

This matters beyond tidiness — the two paths have to stay in lockstep or a bfcache-restored page will diverge from a freshly-loaded one. Worth extracting:

```ts
async function applySessionPayload(payload: SessionInitPayload) {
  const { configs, session, user } = payload;
  setupRequired = Boolean(payload.setupRequired);
  _setConfig(configs);

  const config = getClientConfig();
  if (config) {
    config.setAuthToken(session.authToken);
  } else {
    setLocalStorageSession(session);
  }

  useSessionStore.getState().setUser(parseUser(user));
}
```

Then both `initSession` and `revalidateSessionOnPageShow` call it.

omegascorp (changes requested):
Thanks for this — the core fix is right, and catching the stale-session hole that bfcache opens up (log out elsewhere, hit Back, see an authenticated UI) is good work that the original issue didn't ask for.

I checked out the branch: the three affected test files pass (26 tests) and typecheck is clean against baseline.

Two changes requested, both in `session.ts` — see inline comments. Beyond those, a few non-blocking notes:

- The new `session.ts -> query.ts` import isn't a cycle today, but it couples two modules that were previously independent. An emitter or a callback registered by `renderApp` would keep them decoupled. Fine to leave if you prefer the directness.
- The `pageshow` listener is never removed. No leak in practice since `renderApp` runs once per page, just asymmetric with `disconnectModelenceQueryClient` existing as a teardown path.
- `revalidateModelenceQueries` isn't exported from `client.ts` while its siblings are — presumably intentional, just flagging.
- Please update the PR description: it only covers the `unload` removal, but the `pageshow`/revalidation work is the larger and riskier half and should be reflected in the changelog.
- Test gaps: the new session test doesn't cover the `catch` branch, nor the `getClientConfig()` truthy branch (`config.setAuthToken` is never asserted — only the `setLocalStorageSession` fallback). The second is the path most apps actually take.

ankit-0903:
@omegascorp Thanks for the review!
I've updated the PR with both requested changes:

   - *Extracted applySessionPayload:* Unified session initialization logic
   across initSession and revalidateSessionOnPageShow.
   - *Heartbeat Re-arm:* revalidateSessionOnPageShow now resets and re-arms
   loopSessionHeartbeat() on BFCache restore.
   - *Added Unit Tests: *Covered custom getClientConfig().setAuthToken(...)
   handling and error logging branches in session.test.ts.

All 15 unit tests, ESLint, and Prettier checks pass cleanly!

omegascorp (changes requested):
Thanks for the updates — both changes I asked for landed correctly. `applySessionPayload` is now shared by `initSession` and `revalidateSessionOnPageShow`, the heartbeat re-arm is exactly right, and the two test gaps I flagged (the `getClientConfig()` truthy branch and the `catch` branch) are both covered. PR description is updated too.

I checked out the branch at 868d107: full suite passes (893 tests) and typecheck is clean against a `main` baseline.

One blocker left.

### The regression test the issue actually asks for is still missing

Issue #308 asks for "a regression test for the logout → Back → restored-page path asserting the user is anonymous." That path is the entire reason the `unload` listener existed, and it's the one case the new tests don't cover — all three assert the *populated-user* direction (`user?.id` becomes `'99'`, `'100'`).

Nothing asserts that a payload with `user: null` **clears** an already-populated store. That's worth pinning rather than assuming: `parseUser` returns `null` from two different routes — the `!user` early return, and the `safeParse` failure path after logging — so the anonymous outcome has more than one way in and no test holds either of them.

Please add a test that seeds a logged-in store, revalidates with `user: null`, and asserts `useSessionStore.getState().user` is `null`. That's the logout → Back → restore case in a unit test, and it's what makes this PR safe to merge.

### On the two Bugbot findings — both can be dismissed

I don't want you chasing these, so to be explicit:

**"Live queries stale after BFCache" is wrong as written.** In `packages/modelence/src/websocket/socketio/client.ts`, socket.io auto-reconnects and the `connect` handler calls `resubscribeAll()`, which re-emits every entry in `activeLiveSubscriptions` — and it reads `getAuthToken()` fresh at that moment, so re-subscription carries the *new* token. The subscription is re-established, not left waiting on a dead socket.

**"Stale cache survives BFCache restore"** is largely answered by `invalidateQueries()`, which marks everything stale and refetches active observers regardless of the 60s `staleTime`.

There is a real residue, but it's pre-existing and not yours to fix here: the socket handshake `auth: { token }` in `init()` is captured once and never refreshed, so the handshake identity after a reconnect is the pre-logout token even though the per-subscription payloads carry the new one. I'll track that separately.

### Non-blocking

- The last checklist item in #308 — re-evaluating the `Cache-Control: no-store` headers in `viteServer.ts` — isn't touched here. It's dev-server-only so it doesn't affect production bfcache, and I'm fine deferring it; just don't let the issue auto-close as fully done on merge.
- `revalidateModelenceQueries` still isn't exported from `client.ts` while `connectModelenceQueryClient` / `disconnectModelenceQueryClient` are.
- My earlier notes on the `session.ts -> query.ts` coupling and the never-removed `pageshow` listener still stand. Both are defensible as-is.

Add the anonymous-path test and this is good to merge.

omegascorp (changes requested):
Thanks for this — removing the empty `unload` listener is the right fix for #308, and the session-revalidation half is a sensible companion with good test coverage. A few things should be addressed before merge.

## Blocking

**1. Heartbeat loops can multiply on restore**

`revalidateSessionOnPageShow` clears `heartbeatTimer` and sets it to `null`, then awaits the network call. During that await window `heartbeatTimer === null`, so a concurrent `startSessionHeartbeat()` — or a second `pageshow`, which iOS fires repeatedly on back/forward — passes its `heartbeatTimer !== null` guard and starts a second loop. The `finally` then starts another. Each loop overwrites `heartbeatTimer`, so the orphaned ones become unstoppable: `stopHeartbeatTimer()` can only cancel the last one. Please guard with an explicit "revalidation in flight" flag, or re-check before re-arming in `finally`.

**2. Blanket `invalidateQueries()` also invalidates live queries**

`modelenceLiveQuery` sets `staleTime: Infinity` and all `refetchOn*: false` precisely because it is socket-driven. An unfiltered `invalidateQueries()` marks those stale and refetches the active ones, re-running the promise-based `queryFn` and re-adding resolvers to a subscription that may never fire again — data only resolves on the *next* socket push. Suggest scoping it:

```ts
void queryClientRef.invalidateQueries({ predicate: (q) => q.queryKey[0] !== 'live' });
```

## Worth resolving now

**3. New circular import: `session.ts` → `query.ts` → `method.ts` / `websocket/client`**

`session.ts` importing `./query` pulls the websocket/socket.io client graph into the session module graph, which previously stayed lazy for apps that never call `modelenceLiveQuery`. Note `session.test.ts` mocks `./query`, so the tests won't catch a cycle. Cleaner to invert the dependency: have `renderApp` call `revalidateModelenceQueries()` after `revalidateSessionOnPageShow()` resolves, keeping `session.ts` free of the query dependency.

**4. `applySessionPayload` drops the reconciliation guard**

`hydrateSession` deliberately preserves a localStorage token when the server responds anonymously ("overwriting here would log the user out permanently"). `applySessionPayload` has no equivalent guard, so on restore an anonymous `_system.session.init` response overwrites the stored session with the anon token. That may well be correct for `pageshow` — the user really did log out — but it's the exact scenario the existing comment warns about, and the divergence isn't explained. Please add a comment stating why the paths differ.

**5. `setupRequired` is re-set on restore but nothing re-reads it**

`applySessionPayload` writes `setupRequired`, but `AppProvider`'s `needsSetup` is only updated inside the one-shot `useEffect` guarded by module-level `isInitialized`. A restore that flips `setupRequired` won't surface. Dev-only and minor, but the shared helper makes it look handled when it isn't.

## Minor

- `query.test.ts` calls `connectModelenceQueryClient` with a fake client and never disconnects, leaking `queryClientRef` into later tests in that file. Please add a cleanup.
- The PR title and description still only describe removing the `unload` listener. Please mention the session-revalidation change, since that's where the behavioral risk sits.

## Follow-up (not blocking)

WebSockets aren't re-established on BFCache restore — `websocketsStarted` stays `true` and `startWebsockets()` warns-and-returns, so live queries silently stop updating after a back-nav. Not a regression from this PR (the `unload` listener meant the path never existed), but this PR is what makes it reachable. Tracked separately.

omegascorp (approved):
Thank you, looks good. I'm not merging the PR now, since I'm working on the 0.24.0 release, I will merge it once it's done

ankit-0903:
Thanks @omegascorp! Sounds good.

PR #327, opened 2026-07-25
fix: remove redundant fallback for startTs in cron jobs

### LINKED ISSUE #326: Redundant fallback logic in handleCronJobCompletion
Hey team! 👋 I noticed there's a small piece of redundant fallback logic in `handleCronJobCompletion`. 

Right now, the code falls back to `Date.now()` if `state.startTs` is missing (`state.startTs ? state.startTs + params.interval : Date.now()`). However, `state.startTs` is actually hardcoded to be set immediately at the top of `runCronJob`, which means this fallback is technically dead code. 

It would be great to clean this up for clarity. 😊

aadityakumarsah:
Fixes #326

### What does this PR do?
This PR removes the dead fallback logic in `handleCronJobCompletion`.

### Why are we doing this?
In the current implementation, `state.startTs` is guaranteed to be set at the very top of `runCronJob`. Therefore, checking for its existence and providing a fallback to `Date.now()` in `handleCronJobCompletion` is unnecessary and technically dead code. 

I've replaced it with a non-null assertion to simplify the code and remove the redundant fallback. Let me know what you think! 🚀

---

> [!NOTE]
> **Medium Risk**
> Profile updates are newly throttled (behavior change for heavy updaters), and cron scheduling logic changed in a concurrency-sensitive path—both are important but bounded with tests.
> 
> **Overview**
> Adds **per-user rate limiting** for profile updates via a new `updateProfile` auth bucket (defaults: 30 per 15 minutes and 200 per day), configurable through `auth.rateLimits.updateProfile`. `handleUpdateProfile` calls `consumeRateLimit` only after field validation, the optional `validateProfileUpdate` hook, and handle-uniqueness checks—invalid or rejected updates do not burn quota.
> 
> Separately, **cron job completion** is hardened: each run captures `currentStartTs` and `handleCronJobCompletion` no-ops if `state.startTs` no longer matches, so a slow or overlapping run cannot reschedule or clear state for a newer execution. Scheduling uses `currentStartTs + interval` instead of a `Date.now()` fallback on `startTs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f373c6952127c2d0d8897390174e7ec3a4d44fb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/authConfig.ts
+++ b/packages/modelence/src/app/authConfig.ts
@@ -85,6 +85,8 @@ export type AuthRateLimitsConfig = {
   magicLink?: AuthRateLimitOverride[];
   /** Rate limits for one-time code sign-in attempts. */
   oneTimeCode?: AuthRateLimitOverride[];
+  /** Per-user rate limits for profile updates. */
+  updateProfile?: AuthRateLimitOverride[];
 };
 
 type GenerateHandleProps = {
--- a/packages/modelence/src/auth/profile.test.ts
+++ b/packages/modelence/src/auth/profile.test.ts
@@ -1,6 +1,58 @@
-import { getOwnProfile } from './profile';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mockRequireById = vi.fn();
+const mockFindOne = vi.fn();
+const mockUpdateOne = vi.fn();
+const mockValidateProfileFields = vi.fn();
+const mockValidateProfileUpdate = vi.fn();
+const mockConsumeRateLimit = vi.fn();
+
+vi.doMock('./db', () => ({
+  usersCollection: {
+    requireById: mockRequireById,
+    findOne: mockFindOne,
+    updateOne: mockUpdateOne,
+  },
+}));
+
+vi.doMock('./validators', () => ({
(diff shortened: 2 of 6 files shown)


omegascorp:
@aadityakumarsah this PR contains extra changes, please review commits from other branches

omegascorp:
Closing this one. Thanks for the work here, but it can't be merged in its current form:

- The first commit (profile-update rate limiting) already landed separately via #313.
- The branch now conflicts with `main` — #345 moved the `lastStartDate` upsert inside the `try` block in `runCronJob`, which overlaps this diff.
- The premise of #326 isn't quite right: the `Date.now()` fallback is reachable. When a job times out, `tickCronJobs` clears `isRunning` but leaves `startTs` set, so the job reruns while the previous handler is still running. When the old handler finishes it clears `startTs`, and the newer run then hits the fallback.

That said, the `currentStartTs` guard in your last commit is a real fix for that overlapping-run race and is worth landing. If you'd like to pick it back up, please open a fresh PR rebased on `main` containing only that change, plus a test in `jobs.test.ts` covering the timeout/overlap case.

PR #328, opened 2026-07-26
feat: add private fields in schema

AtulYadav25:
#  PR Overview

### 1. Added`.private()` Schema Support & Dot Notation
Added`isFieldPrivate()` and `extractPrivateFieldPaths()` to recursively extract private field paths from nested `z.ZodObject`, `z.ZodArray`, `z.ZodOptional`, `z.ZodNullable`, and `z.ZodEffects`. Added `SelectKey<TSchema>` to provide clean IDE autocompletion for top-level and partially dot-notation paths.

Example:
```ts
import { Store, schema } from 'modelence/server';

export const dbTodos = new Store('todos', {
  schema: {
    title: schema.string(),
    completed: schema.boolean(),
    userId: schema.string().private(), //Private field, automatically extracted when document is fetched
    groupInfo: schema.object({
      users: schema.array(schema.string()).private(),  //Private field, automatically extracted when document is fetched
      wager: schema.number(),
      platform: schema.string()
    })
  },
  indexes: []
});

```

### 2. Added `select` Option to `Store` Read Operations
Added `select` to `FindOptionsWithSelect` and `FetchOptionsWithSelect` across `findOne`, `requireOne`, `findById`, `requireById`, and `fetch`. This allows developers to un-hide specific private fields on demand while keeping all other private fields hidden.

***Partial Type Safe***  (Related to Schema declared above)
Screenshot (848)

***Type Safe Top Level and Nested Fields too***  (Related to Schema declared above)
typesafe for select

### 3. Improved `projection` Handling (Compile-Time & Runtime)
- **Compile-Time (`FetchedDoc`):**
  - **Inclusion Projections (`projection: { apiKey: 1 }`):** Private fields included in the projection become **visible** in the return type.
  - **Exclusion Projections (`projection: { active: 0 }`):** Private fields not excluded remain **visible** in the return type.
  - **Private Field Exclusions (`projection: { apiKey: 0 }`):** Excluded private fields remain **hidden**, while other private fields become **visible**.

- **Runtime:**
  - `select` and `projection` are merged together instead of `projection` overriding `select`.
---

### Scenario 1: Schema Definition with Top-Level and Nested Private Fields

```ts
export const organizationStore = new Store('organizations', {
  schema: {
    name: schema.string(),
    active: schema.boolean(),
    apiKey: schema.string().private(), // Top-level private field
    teamSettings: schema.object({
      memberEmails: schema.array(schema.string()).private(), // Nested private field
      maxQuota: schema.number(),
    }),
  },
  indexes: [],
});
```

### Scenario 2: Un-hiding Private Fields via `select` (Added in this PR)

```ts
// Un-hide nested private field using dot notation
const orgs = await organizationStore.fetch({}, {
  select: ['teamSettings.memberEmails']
});

// ✅ IDE provides autocompletion for 'teamSettings.memberEmails'
// ✅ orgs[0].teamSettings.memberEmails is typed as string[]
```

### Scenario 3: Un-hiding Private Fields via Inclusion `projection` (Improved in this PR)

```ts
// Previously on main: doc.apiKey was typed as undefined
// Now on this branch: doc.apiKey is typed as string
const doc = await organizationStore.findOne(
  { name: 'Acme Corp' },
  { projection: { name: 1, apiKey: 1 } }
);

console.log(doc?.apiKey); // ✅ Typed as string
```

### Scenario 4: Un-hiding Private Fields via Exclusion `projection` (Improved in this PR)

```ts
// 4a. Excluding a non-private field (active: 0):
// MongoDB returns all other fields, so all private fields are un-hidden:
const doc1 = await organizationStore.findOne(
  { name: 'Acme Corp' },
  { projection: { active: 0 } }
);
console.log(doc1?.apiKey); // ✅ Un-hidden (string)
console.log(doc1?.teamSettings.memberEmails); // ✅ Un-hidden (string[])

// 4b. Excluding a specific private field (apiKey: 0):
const doc2 = await organizationStore.findOne(
  { name: 'Acme Corp' },
  { projection: { apiKey: 0 } }
);
console.log(doc2?.apiKey); // ✅ Hidden (undefined)
console.log(doc2?.teamSettings.memberEmails); // ✅ Un-hidden (string[])
```

### Scenario 5: Dual `projection` + `select` Runtime Merging (Added in this PR)

```ts
await organizationStore.fetch(
  { name: 'Acme Corp' },
  { projection: { name: 1 }, select: ['apiKey'] }
);

// ✅ MongoDB query sent: { projection: { name: 1, apiKey: 1 } }
// ✅ Return type: apiKey is un-hidden
```

| Feature / Behavior | Current Feature Branch |
| :--- | :--- |
| **Private Fields (`.private()`)** | ✅ **Full recursive support** for top-level and deeply nested fields  |
| **`select` Option in Read Operations** | ✅ **Added `select` option** (`select: ['apiKey', 'groupInfo.users']`) across all store read methods. |
| **Dot Notation Support** | ☑ ** Partial Type Safe (`SelectKey<TSchema>`)** for top-level keys and nested dot paths (e.g. `'teamSettings.memberEmails'`). |
| **`projection` Return Types** | ✅ **Improved `projection` typing:** Document return types automatically un-hide private fields included in **Inclusion** (`{ apiKey: 1 }`) or **Exclusion** (`{ active: 0 }`) projections. |
| **`projection` + `select` Interaction** | ✅ **Runtime Merging:** `select` keys are intelligently merged into `projection` objects before querying MongoDB. |
| **`projection` Key Safety** | ✅ **Strict Compile-Time Validation:** Invalid top-level schema keys in `projection` are rejected by TypeScript. |

---

> [!NOTE]
> **High Risk**
> Changes default read/mutation return shapes and runtime field stripping for secrets—incorrect stripping or select/projection edge cases could leak or drop data across many Store methods.
> 
> **Overview**
> Adds **`.private()`** on schema fields so sensitive data is **stripped by default** on store reads, with optional **`select`** (including dot paths) and smarter **`projection`** to reveal only what you intend.
> 
> **Schema & types:** Zod gets a `.private()` marker (registry + unwrapping through optional/nullable/branded/etc.), `extractPrivateFieldPaths`, serialized `private: true` in API schema output, and helpers **`inferFetched` / `inferSelected`** (docs nav updated).
> 
> **Store behavior:** Read and atomic-return paths (`findOne`, `fetch`, `findById`, `require*`, `findOneAnd*`) now return **`FetchedDoc`** types that omit private fields unless `select` or an **inclusion** projection includes them; runtime **`stripPrivateFields`** uses new **`projectionHelpers`** to delete nested paths without breaking parent containers or leaking via bogus sub-paths on primitives/Dates. **`select` is stripped** before MongoDB; inclusion **`projection` + `select` merge** avoids path collisions. **`requireById` / `requireOne`** accept options or a custom error handler as the second argument.
> 
> **Tests:** Extensive coverage for nested/private-parent cases, arrays, branded fields, `$slice`, and mutation methods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8661b7e1b9d6fe7f680f409923d8498326a7cd33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AtulYadav25:
@omegascorp  PR is ready for review.

I'm still stress testing these changes, but it is open for review.
Let me know if you want any changes.

AtulYadav25:
I'm still testing, and this also needs documentation update. I think it's better to open another PR dedicated for docs

Ready for review

omegascorp:
Hi @AtulYadav25 could you check the following, it's generated by AI, but focused on the points, I've asked to highlight

1. Architecture
The core shape is right: fetch the full doc, strip at the boundary (wrapFetchedDocument → stripPrivateFields), with private paths computed once in the constructor. That's the correct call over building Mongo projections, because projection-based hiding breaks down on $slice/$elemMatch and nested arrays. Splitting the traversal primitives into [projectionHelpers.ts](vscode-webview://152mmo9b5a5plsc66ua13shcn0ei5bnphvrvr6i7fgluigvvo704/packages/modelence/src/data/projectionHelpers.ts) is good, and the type layer in isolation is correct — I asserted that InferFetchedDocumentType hides apiKey and InferSelectedDocumentType<_, 'apiKey'> reveals it.

Real issues, in severity order:

a) .private() mutates a shared Zod instance — privacy leaks across fields. [types.ts:47](vscode-webview://152mmo9b5a5plsc66ua13shcn0ei5bnphvrvr6i7fgluigvvo704/packages/modelence/src/data/types.ts#L47) does privateSchemaRegistry.add(this); return this. Since .private() returns the same object, any reused schema instance becomes private everywhere. Confirmed:


const shared = schema.string();
const a = { secret: shared.private() };
const b = { publicName: shared };   // never marked private
extractPrivateFieldPaths(b)  // → ['publicName']   ← wrong
Same for a reused schema.object(...): marking one field private marked both. This is the inverse of a leak (over-hiding), so it silently drops data rather than exposing it — but it's still a correctness bug, and commit 69f8806 ("fix: Private marker mutates schema") did not actually fix it. Fix: return a clone, e.g. const copy = new (this.constructor as any)(this._def); registry.add(copy); return copy; — or drop the WeakSet and carry the marker in _def.

b) $slice is classified as an inclusion projection. [projectionHelpers.ts:8](vscode-webview://152mmo9b5a5plsc66ua13shcn0ei5bnphvrvr6i7fgluigvvo704/packages/modelence/src/data/projectionHelpers.ts#L8) treats any object value as inclusion. In MongoDB, { comments: { $slice: 1 } } is neither — it returns all other fields. Runtime happens to stay safe (I verified apiKey is still stripped), because stripPrivateFields only un-hides names in the projection. But getMongoProjection uses the same predicate to decide whether to merge select keys, so { projection: { comments: { $slice: 1 } }, select: ['apiKey'] } builds a mixed projection Mongo will reject. $slice/$meta/$elemMatch should be their own category, not inclusion.

c) The type-level privacy guarantee is defeated by TMethods. On the fetched doc, apiKey doesn't resolve to string or to absent — it resolves to a method signature:


Type '(this: ..., ...args: any[]) => any' is not assignable to type 'string'
Because TMethods extends Record<string, (...) => any> ([store.ts:541](vscode-webview://152mmo9b5a5plsc66ua13shcn0ei5bnphvrvr6i7fgluigvvo704/packages/modelence/src/data/store.ts#L541)) has an open string index signature, keyof FetchedDoc includes every string. So 'apiKey' extends keyof Fetched is true even though the mapped type omitted it. This is pre-existing on main, not introduced here — but it's load-bearing for this PR: the headline "private fields are hidden in the return type" doesn't hold end-to-end until TMethods defaults to something without an index signature. Worth fixing in or alongside this PR, since it's the feature's main selling point.

d) stripPrivateFields mutates the driver's document in place. Fine for findOne, but it's a side effect on an object the caller may hold. Given wrapDocument already builds a new object when methods exist, stripping into a copy would be more consistent with the codebase's immutability convention.

e) Coverage gaps. aggregate(), watch(), distinct(), and create() all return unstripped data. create() echoing back the input is defensible, but aggregate() is a plausible leak path. At minimum document that privacy is a find-family guarantee only.

2. Can it be simplified?
Yes — substantially. ~2278 added lines for this feature is roughly 2–3× what's needed.

Biggest win: drop the projection-driven type gymnastics (~150 lines). IsInclusionValue, IsExclusionValue, IsInclusionProjection, ExtractInclusionKeys, ExtractExclusionKeys, ExtractProjectionVisibleKeys, OmitExclusions, and the KProjection generic threaded through nine method signatures exist to make { apiKey: 1 } un-hide apiKey in the return type. But select already does that, explicitly and legibly. Making projections also a privacy-unhiding channel means two mechanisms for one job, and it's the source of the $slice inconsistency in (b). Keep projection as a pure Mongo pass-through, keep select as the single privacy escape hatch, and every KProjection generic disappears. This alone likely halves the diff.

Collapse the three near-identical options types. FindOptionsWithSelect, FetchOptionsWithSelect, and FindOneAndModifyOptionsWithSelect repeat the same select?: KSelect[] + projection-key-validation block. One generic WithSelect<TBase, TSchema, KSelect> covers all three.

UnwrapTransparent + InferSelectedZodType duplicate the same unwrap list three times — once at the type level in UnwrapTransparent, once in InferSelectedZodType's conditional chain, once at runtime in extractPrivateFieldPaths. The runtime one can't merge, but the two type-level chains can share a single unwrap helper.

normalizeRequireOptions overloads ([store.ts:1147](vscode-webview://152mmo9b5a5plsc66ua13shcn0ei5bnphvrvr6i7fgluigvvo704/packages/modelence/src/data/store.ts#L1147)) add a positional-polymorphism API (options | (() => Error) in slot 2) to preserve backwards compat on requireOne/requireById. It works, but if you're pre-1.0 it's cheaper to just make the errorHandler an options.onNotFound field and take the breaking change.

isFieldPrivate and extractPrivateFieldPaths each hand-roll the unwrap ladder. extractPrivateFieldPaths already recurses through every wrapper — isFieldPrivate could be a registry lookup plus one shared unwrapOnce() helper rather than eight instanceof branches.

artahian:
@omegascorp what does `.private()` mean in general? This seems too specific for this particular user object case. The schema should have a general meaning for all collections, and `private` doesn't really fit anything because it sounds more like a Supabase-style row/field level security gate.

Modelence's direction is more logic based - every query and mutation decides what and how it should fetch. I don't think we need a blanket rule at a single field level (i.e. adding `private` to the whole schema just for a particular user's case).

omegascorp:
@artahian The generic idea was to strip private fields during serialization so they never transferred to the frontend. Though the PR is quite big I will need to review again what can be simplified here.

packages/modelence/src/data/store.ts
```diff
@@ -394,14 +551,23 @@ export class Store<
   /** @internal */
   readonly _doc!: this['_rawDoc'] & TMethods;
 
+  /** @internal */
+  readonly _fetchedType!: InferFetchedDocumentType<TSchema>;
+  /** @internal */
+  readonly _fetchedRawDoc!: WithId<this['_fetchedType']>;
+  /** @internal */
+  readonly _fetchedDoc!: this['_fetchedRawDoc'] & TMethods;
```
omegascorp, line 559: **Type-level hiding does not work for stores declared without `methods`.**

`_fetchedDoc` intersects with `TMethods`, and `TMethods` has no default — so when `methods` is omitted, TS infers it from its own constraint, producing an index-signature-ish record that swallows every key. Verified against this branch:

```ts
const s = new Store("t", { schema: { apiKey: schema.string().private() }, indexes: [] });
const a = await s.findOne({});
a?.apiKey;          // compiles — should be hidden
a?.totallyBogusKey; // also compiles — proves the index-signature swallow
```

A store declared *with* `methods` hides correctly; without it, nothing is hidden. Since most stores omit `methods`, the compile-time half of this feature is off for the common case.

To be fair: this shape is **pre-existing on `main`** (the bogus-key probe fails there too), so the PR did not introduce it. But type-level hiding is this PR's entire value proposition, so it can't ship as advertised without a `TMethods = Record<string, never>` default. Worth confirming that default does not break inference for stores that do use methods.

Runtime stripping is unaffected — this is a false sense of compile-time safety, not a data leak.

packages/modelence/src/data/types.ts
```diff
@@ -28,6 +28,136 @@ const schemaObject: typeof z.object = z.object.bind(z);
...
+
+/**
+ * Registry of Zod schema instances that have been marked as private via `.private()`.
+ *
+ * The WeakSet holds no strong references, so schemas are
+ * garbage-collected normally when no longer in use.
+ */
+const privateSchemaRegistry = new WeakSet<z.ZodType>();
+
+(z.ZodType.prototype as any).private = function (this: z.ZodType) {
```
omegascorp, line 48: **`.private()` mutates shared schema instances.**

This registers `this` in the WeakSet and returns the same object, so a reused schema constant becomes private everywhere it is referenced. Verified:

```ts
const str = schema.string();
const s = { pub: str, priv: str.private() }; // BOTH become private
```

Returning a clone instead of `this` would make this safe. Low likelihood, but silent and confusing when it happens.

omegascorp (commented):
Review notes on the type layer and schema traversal. Runtime stripping and test coverage look solid — I ran the suite (172 pass) and verified the behavior below empirically rather than by inspection. Details inline.

packages/modelence/src/data/types.ts
```diff
@@ -28,6 +28,136 @@ const schemaObject: typeof z.object = z.object.bind(z);
...
+  } else if (schemaDef instanceof z.ZodOptional) {
+    paths.push(...extractPrivateFieldPaths(schemaDef.unwrap(), prefix));
+  } else if (schemaDef instanceof z.ZodNullable) {
+    paths.push(...extractPrivateFieldPaths(schemaDef.unwrap(), prefix));
+  } else if (schemaDef instanceof z.ZodDefault) {
+    paths.push(...extractPrivateFieldPaths(schemaDef.removeDefault(), prefix));
+  } else if (schemaDef instanceof z.ZodEffects) {
+    // ZodEffects (transform/refine) has no public unwrap — _def.schema is the only accessor.
+    paths.push(...extractPrivateFieldPaths((schemaDef._def as any).schema, prefix));
+  } else if (schemaDef instanceof z.ZodUnion) {
```
omegascorp, line 105: **`extractPrivateFieldPaths` silently misses four Zod constructs.**

Plain `z.union` is handled here, but these each return `[]` instead of the private path — confirmed by test on this branch:

- `z.record(z.string().private())`
- `z.discriminatedUnion(...)` branches
- `z.lazy(...)` / recursive schemas
- `z.tuple(...)`

The `union` vs `discriminatedUnion` asymmetry looks accidental, and `discriminatedUnion` is the most likely of these to appear in real schemas.

Silent misses are the dangerous failure mode for a security-adjacent feature: a user writes `.private()`, gets no error, and the field ships unstripped. Either support these, or throw at Store construction when `.private()` sits under an unsupported wrapper.

omegascorp:
Two more items that do not map to a single changed line:

**1. The PR description contradicts the implementation on exclusion projections.**

Scenario 4a claims `projection: { active: 0 }` un-hides `apiKey`. It does not — commit 75e45ca deliberately changed this so exclusions keep private fields hidden. Verified at runtime on this branch: `{ name: 0 }` returns the doc with `secret` **removed**.

The code is right; the description is stale and describes leakier behavior than what actually ships. Worth fixing before merge so reviewers and users are not misled about the security posture.

**2. Simplification: collapse the duplicated projection-validation types.**

`FindOptionsWithSelect`, `FetchOptionsWithSelect`, and `FindOneAndModifyOptionsWithSelect` each repeat a near-identical block:

```ts
projection?: KProjection & TypedProjection<...> & {
  [K in keyof KProjection]: K extends ... ? KProjection[K] : never;
};
```

These should collapse into a single `ValidatedProjection<TDoc, KProjection>` helper. `FindOptionsWithSelect` and `FindOneAndModifyOptionsWithSelect` are otherwise the same type modulo the base options, so they can likely merge outright.

Minor, non-blocking: `IsInclusionValue` (type) and `isProjectionInclusionValue` (runtime) encode the same rule twice with different logic — the type treats bare `number`/`boolean` as inclusion while `IsExclusionValue` special-cases them. That divergence is subtle enough to deserve a comment. Same for `isContainer` returning `true` for `[]` but recursing otherwise.

`projectionHelpers.ts` is clean and well-tested — good extraction.

---

Overall: architecture is sound. Fetch-full-doc-then-strip is the right call, and moving off the projection-building approach was correct since projections alone cannot handle private fields inside arrays or exclusion projections. The blocker is the type layer, not the runtime.

omegascorp:
Closing this for now. Thank you for the substantial effort here — the runtime approach (fetch the full document, strip at the boundary) is the right shape, and the test coverage is solid.

Why we're closing rather than iterating further:

**1. The design direction is still open.** As @artahian raised above, a field-level `.private()` flag doesn't map cleanly onto Modelence's query-level authorization model, and we haven't settled whether serialization-time stripping is something the framework should own. That's a decision we need to make before landing a change of this size to `Store`.

**2. Blocking issues from the last review round remain unaddressed** (no activity since Aug 5):
- Type-level hiding is defeated for any store declared without `methods`, which is most stores.
- `.private()` still mutates the shared Zod instance, so a reused schema constant becomes private everywhere. Commit 69f8806 didn't resolve this.
- `extractPrivateFieldPaths` silently misses `z.record`, `z.discriminatedUnion`, `z.lazy`, and `z.tuple`. For a security-adjacent feature, silent misses are the dangerous failure mode.
- `$slice` is treated as an inclusion projection, and combining it with `select` builds a mixed projection Mongo rejects.
- `aggregate()`, `watch()`, and `distinct()` return unstripped data.

**3. The diff is roughly 2–3× larger than the feature needs.** The projection-driven type gymnastics (`IsInclusionValue`, `ExtractProjectionVisibleKeys`, the `KProjection` generic threaded through nine signatures) duplicate what `select` already does. If we pick this back up, `select` should be the single unhide mechanism and `projection` a pure Mongo pass-through.

If the team lands on wanting this, a fresh, smaller PR is welcome — starting with the `TMethods` default fix, since the compile-time guarantee depends on it. We'll link the design discussion here when there's a conclusion.

PR #329, opened 2026-07-28
fix: prevent memory leak from process event listeners

aadityakumarsah:
This PR fixes a MaxListenersExceededWarning caused by repeatedly registering `process.on` listeners in `startServer` across tests. It uses a module-level variable to ensure they are added only once.

---

> [!NOTE]
> **Low Risk**
> Small control-flow change around existing logging handlers; no new behavior on first boot and no auth or data-path impact.
> 
> **Overview**
> **Fixes listener accumulation** when `startServer` runs multiple times in the same Node process (e.g. repeated test runs), which triggered `MaxListenersExceededWarning`.
> 
> Global `unhandledRejection` and `uncaughtException` handlers are now registered behind a module-level `globalProcessListenersRegistered` guard so they attach **once**; later `startServer` calls skip re-registration. Handler behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e6c1dd3dd4834edffbf41222d70d75058612230. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #330, opened 2026-07-29
fix(cron): remove redundant fallback logic in handleCronJobCompletion

### LINKED ISSUE #326: Redundant fallback logic in handleCronJobCompletion

aadityakumarsah:
Resolves #326. Removed dead fallback logic since startTs is always set.

---

> [!NOTE]
> **Low Risk**
> Small behavioral hardening and scheduling cleanup in cron and server bootstrap with no auth, data, or API contract changes.
> 
> **Overview**
> **Cron:** `handleCronJobCompletion` now sets the next `scheduledRunTs` with `(state.startTs ?? Date.now()) + params.interval` instead of a branch that used `Date.now()` only when `startTs` was missing—aligning with `startTs` always being set when a job finishes via `runCronJob`.
> 
> **Server:** `unhandledRejection` and `uncaughtException` listeners are registered behind a module-level `globalProcessListenersRegistered` flag so multiple `startServer` invocations (e.g. in tests) do not attach duplicate handlers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e8770ab2fc9df948ace7781ca44ff8ceb1b93c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/server.ts
+++ b/packages/modelence/src/app/server.ts
@@ -75,6 +75,8 @@ function registerModuleRoutes(app: express.Application, modules: Module[]) {
   }
 }
 
+let globalProcessListenersRegistered = false;
+
 export async function startServer(
   server: AppServer,
   {
@@ -161,18 +163,21 @@ export async function startServer(
     Promise.resolve(server.handler(req, res)).catch(next);
   });
 
-  process.on('unhandledRejection', (reason, promise) => {
-    console.error('Unhandled Promise Rejection:');
-    console.error(reason instanceof Error ? reason.stack : reason);
-    console.error('Promise:', promise);
-  });
+  if (!globalProcessListenersRegistered) {
+    globalProcessListenersRegistered = true;
+    process.on('unhandledRejection', (reason, promise) => {
+      console.error('Unhandled Promise Rejection:');
+      console.error(reason instanceof Error ? reason.stack : reason);
+      console.error('Promise:', promise);
+    });
 
-  // Global uncaught exceptions
-  process.on('uncaughtException', (error) => {
-    console.error('Uncaught Exception:');
-    console.error(error.stack); // This gives you the full stack trace
-    console.trace('Full application stack:'); // Additional context
-  });
+    // Global uncaught exceptions
+    process.on('uncaughtException', (error) => {
+      console.error('Uncaught Exception:');
+      console.error(error.stack); // This gives you the full stack trace
+      console.trace('Full application stack:'); // Additional context
+    });
+  }
 
   const websocketProvider = getWebsocketConfig()?.provider;
   if (websocketProvider) {
--- a/packages/modelence/src/cron/jobs.ts
+++ b/packages/modelence/src/cron/jobs.ts
@@ -153,7 +153,7 @@ async function runCronJob(job: CronJob) {
 }
 
 function handleCronJobCompletion(state: CronJob['state'], params: CronJob['params']) {
-  state.scheduledRunTs = state.startTs ? state.startTs + params.interval : Date.now();
+  state.scheduledRunTs = state.startTs! + params.interval;
   state.startTs = undefined;
   state.isRunning = false;
 }
--- a/packages/modelence/src/cron/jobs.ts
+++ b/packages/modelence/src/cron/jobs.ts
@@ -153,7 +153,7 @@ async function runCronJob(job: CronJob) {
 }
 
 function handleCronJobCompletion(state: CronJob['state'], params: CronJob['params']) {
-  state.scheduledRunTs = state.startTs! + params.interval;
+  state.scheduledRunTs = (state.startTs ?? Date.now()) + params.interval;
   state.startTs = undefined;
   state.isRunning = false;
 }


packages/modelence/src/cron/jobs.ts
```diff
@@ -153,7 +153,7 @@ async function runCronJob(job: CronJob) {
 }
 
 function handleCronJobCompletion(state: CronJob['state'], params: CronJob['params']) {
-  state.scheduledRunTs = state.startTs ? state.startTs + params.interval : Date.now();
+  state.scheduledRunTs = (state.startTs ?? Date.now()) + params.interval;
```
omegascorp, line 156: This change is correct, and it's a genuine bug fix — but not the one the PR describes.

The old code was `state.startTs ? state.startTs + params.interval : Date.now()`. Note the fallback branch had **no `+ params.interval`** — so any job reaching it was rescheduled to fire immediately rather than one interval later. Your version fixes that. Good catch, even if it was incidental.

But the fallback isn't dead. Look at the timeout handling in `tickCronJobs`:

```ts
if (state.isRunning) {
  if (state.startTs && state.startTs + params.timeout < now) {
    state.isRunning = false;   // startTs deliberately left set
  }
  return;
}
```

That clears `isRunning` without clearing `startTs`, so the next tick can start a *fresh* run of the same job. When the original (still-pending) handler eventually settles, it calls `handleCronJobCompletion` on state the newer run has already mutated — and `startTs` may well be `undefined` by then. So `??` is doing real work here.

Could you add a test that drives this path? Something that starts a job, forces the timeout branch, lets the original handler resolve, and asserts `scheduledRunTs` lands one interval out rather than at `Date.now()`. Without it, this line is going to get "cleaned up" again by the next reader — which is exactly how we got here.

packages/modelence/src/app/server.ts
```diff
@@ -75,6 +75,8 @@ function registerModuleRoutes(app: express.Application, modules: Module[]) {
   }
 }
 
+let globalProcessListenersRegistered = false;
```
omegascorp, line 78: Could this move to a separate PR? It's an unrelated concern from the cron fix, and it deserves its own review rather than riding along on a one-line change.

On the substance: this flag is set once per module load and **never reset**, so the guard can't be unwound. The stated motivation is that repeated `startServer` calls in tests shouldn't stack listeners — but `app/server.test.ts` calls `startServer` ~25 times in one module instance. After the first, the handlers are permanently pinned, and nothing can remove or re-register them. A test that wants to assert on the rejection handler, or a consumer restarting a server after teardown, silently gets no handlers. That's a latch, not lifecycle management.

Two cleaner directions:

- Hold named function references and return a teardown that calls `process.off(...)`, so the registration is actually reversible.
- Or move registration out of `startServer` entirely into a module-init path that runs once by construction — these are process-global concerns, not per-server ones.

Separately, and not blocking since it predates this PR: an `uncaughtException` handler that logs and lets the process continue leaves Node in an undefined state. If you're touching this code anyway, it's a reasonable moment to consider exiting or re-throwing after logging.

omegascorp (changes requested):
Thanks for digging into this! The cron change is correct and actually fixes a real bug — but it isn't the cleanup #326 described, and I'd like the description corrected and the unrelated `server.ts` change split out before this lands.

**Summary of what I'm asking for:**

1. **The fallback is not dead code.** The timeout path in `tickCronJobs` clears `isRunning` on a timed-out job while leaving `startTs` set; the next tick starts a fresh run, and when the original long-running handler finally settles it calls `handleCronJobCompletion` against state the newer run has since mutated. `startTs` can genuinely be `undefined` there. Your `??` is the right call — but the reasoning in the PR body ("dead fallback logic since startTs is always set") is wrong, and so is the premise of #326.

2. **This PR fixes a different bug than it claims.** The old fallback set `scheduledRunTs = Date.now()` with no `+ params.interval`, so a job hitting that branch rescheduled to run *immediately* instead of one interval out. That's the actual defect being fixed. Please retitle to something like `fix(cron): apply missing interval to scheduledRunTs fallback` and update the body. I'd also suggest closing #326 as invalid rather than "resolved" — otherwise the next person reads it and deletes the fallback for real.

3. **Please split the `server.ts` commit into its own PR.** It's an unrelated process-lifecycle concern, and reviewers of a one-line cron fix aren't primed to scrutinize it. Details in the inline comment.

4. **No test coverage on either change.** `cron/jobs.test.ts` doesn't touch `handleCronJobCompletion`, `startTs`, or `scheduledRunTs`. Given that this exact branch was just misread as dead code, a test pinning the timeout path would be worth having.

One note on my own review: I read the code statically — `git fetch origin fix/issue-326` fails since the branch is on your fork, so I haven't run the suite against it.

omegascorp:
Closing this one as well:

- The process-listener change already merged via #329.
- The remaining one-liner (`(state.startTs ?? Date.now()) + params.interval`) is cosmetic and doesn't address the actual problem. The fallback isn't dead code — it's hit when a timed-out job overlaps with its own rerun, and the real fix is the `currentStartTs` guard from #327's last commit.

See the note on #327 for what a follow-up PR should contain.

PR #335, opened 2026-08-05
fix: add regex validation for handle to allow only alphanumeric chars

### LINKED ISSUE #331: Validation: Enforce alphanumeric and URL-safe characters for user handles
### Description
Currently, `validateHandle` in `packages/modelence/src/auth/validators.ts` only enforces length constraints (`MIN_HANDLE_LENGTH` = 3, `MAX_HANDLE_LENGTH` = 50). It does not validate the character format of the handle. 

As a result, handles can contain spaces, special characters, or symbols (e.g. `user name`, `user@name!`), which can break profile URLs and routing.

### Expected Behavior
Handles should only allow alphanumeric characters, underscores, and hyphens (`/^[a-zA-Z0-9_-]+$/`).

### Proposed Solution
Add a `.regex()` validation step to the handle schema in `validators.ts` to ensure handles are URL-safe and consistent.

ankit-0903:
Fixes #331

### Description
Currently `validateHandle` and `validateProfileFields` in `validators.ts` only check handle min and max length (3 to 50 characters). It does not validate what characters are allowed in the handle.

Because of this, handles can be set with spaces or special characters (like `user name` or `user@123`), which can break profile URL links and routes.

### Changes
1. Added `HANDLE_REGEX = /^[a-zA-Z0-9_-]+$/` in `validators.ts`.
2. Updated `handleSchema` to check against `HANDLE_REGEX` so handles can only contain alphanumeric characters, underscores, and hyphens.
3. Added unit tests in `validators.test.ts` to test invalid handle formats (spaces, special symbols) and valid handles.

All unit tests pass and build is clean.

---

> [!NOTE]
> **Medium Risk**
> Changes auth validation and signup handle generation; legacy carve-out reduces breakage for existing bad handles, but new or changed handles are enforced strictly.
> 
> **Overview**
> Handles are now restricted to letters, numbers, underscores, and hyphens via **`HANDLE_REGEX`** on **`validateHandle`** and profile updates, rejecting values like spaces or `@` that could break profile URLs.
> 
> **`handleUpdateProfile`** skips re-validating **`handle`** when the submitted value matches the stored handle (after trim), so accounts that already have “invalid” handles can still change other profile fields without being blocked.
> 
> Auto-generated handles from email local-parts in **`resolveUniqueHandle`** replace disallowed characters with underscores before length padding, and **`AuthConfig.generateHandle`** docs now state custom handles must match the same rules. Tests cover invalid formats, legacy profile updates, and email-derived handle sanitization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1540234b28a968313bf176984d45b3be33ca8633. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/authConfig.ts
+++ b/packages/modelence/src/app/authConfig.ts
@@ -306,8 +306,10 @@ export type AuthConfig = {
    * Custom handle generator. If provided, overrides the default behavior
    * (which derives the handle from the email local-part). Receives
    * `{ email, firstName?, lastName? }` and returns the desired handle
-   * synchronously or as a `Promise<string>`. If the returned handle collides
-   * with an existing one, Modelence appends a numeric suffix automatically.
+   * synchronously or as a `Promise<string>`. The returned handle must contain
+   * only letters, numbers, underscores, and hyphens (matching `HANDLE_REGEX`).
+   * If the returned handle collides with an existing one, Modelence appends
+   * a numeric suffix automatically.
    */
   generateHandle?: (props: GenerateHandleProps) => Promise<string> | string;
 
--- a/packages/modelence/src/auth/profile.test.ts
+++ b/packages/modelence/src/auth/profile.test.ts
@@ -119,5 +119,19 @@ describe('auth/profile', () => {
       });
       expect(mockUpdateOne).toHaveBeenCalled();
     });
+
+    it('allows legacy users with invalid handle characters to update other fields when handle is unchanged', async () => {
+      mockRequireById.mockResolvedValueOnce({
+        _id: 'legacy-1',
+        handle: 'legacy.user',
+      });
+
+      await handleUpdateProfile(
+        { handle: 'legacy.user', firstName: 'Updated' },
+        authenticatedContext
+      );
+
+      expect(mockValidateProfileFields).toHaveBeenCalledWith({ firstName: 'Updated' });
+    });
   });
 });
--- a/packages/modelence/src/auth/profile.ts
+++ b/packages/modelence/src/auth/profile.ts
@@ -29,7 +29,14 @@ export async function handleUpdateProfile(props: Args, { user }: Context) {
 
   let profile = await usersCollection.requireById(user.id);
 
-  const update = validateProfileFields(props as UpdateProfileProps);
+  const rawProps = props as UpdateProfileProps;
+  const isHandleUnchanged =
+    typeof rawProps.handle === 'string' && rawProps.handle.trim() === profile.handle;
+
+  const { handle: _ignoredHandle, ...propsWithoutHandle } = rawProps;
+  const update = validateProfileFields(
+    isHandleUnchanged ? (propsWithoutHandle as UpdateProfileProps) : rawProps
+  );
 
   await getAuthConfig().validateProfileUpdate?.(update);
 
--- a/packages/modelence/src/auth/utils.test.ts
+++ b/packages/modelence/src/auth/utils.test.ts
@@ -1,0 +1,32 @@
+import { describe, test, expect, vi, beforeEach } from 'vitest';
+import { resolveUniqueHandle } from './utils';
+import { usersCollection } from './db';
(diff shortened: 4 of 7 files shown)


omegascorp:
Thanks for this — the core fix is right, and nice catch on the follow-up commit sanitizing the email local-part in `resolveUniqueHandle`. Without that, `john.doe@example.com` wouldn't have been able to sign up at all, since the generated handle would have failed its own validator. Sanitizing before padding is also the correct order (`..@x.com` → `___` passes; padding first would leave `.._`).

Two things I'd like addressed before merge.

### 1. Existing users with legacy handles can't update their profile

Handles created before this validation existed (e.g. `john.doe`) are still in the DB. Login and session lookups don't validate, so those users can sign in fine — but profile updates break. `validateProfileFields` is `.partial()`, so a firstName-only update is unaffected; the problem is the common UI pattern where the form is populated with current values and POSTs all of them back. A legacy user changing their first name then gets:

```
handle: must contain only letters, numbers, underscores, and hyphens
```

They're blocked from editing unrelated fields until they also change their handle.

The fix is to validate the handle only when it's actually changing. In `handleUpdateProfile`, `profile` is already loaded on the line above for the auth check, so this costs no extra query:

```ts
const rawProps = props as UpdateProfileProps;
const isHandleUnchanged =
  typeof rawProps.handle === 'string' && rawProps.handle.trim() === profile.handle;

const { handle: _ignoredHandle, ...propsWithoutHandle } = rawProps;
const update = validateProfileFields(
  isHandleUnchanged ? (propsWithoutHandle as UpdateProfileProps) : rawProps
);
```

Dropping the field rather than bypassing validation also means the unchanged handle isn't rewritten to Mongo, and the "handle already taken" check below correctly skips it. New handles are still fully validated, so the intent of this PR is preserved.

I tested this locally with cases covering both directions — legacy user can update, changing *to* a new invalid handle is still rejected, and users with valid handles are unaffected. Happy to push the patch and tests to your branch if that's easier.

### 2. Needs a changeset — this is a breaking change

An app with a custom `generateHandle` hook returning something like `first.last` will now fail at signup. Nothing in the docs previously stated the character set was constrained, because it wasn't. That can't be made backwards compatible (the value is generated fresh, so there's no "unchanged" case), so it should be called out explicitly:

- Add a changeset marking this a breaking change.
- Update the `generateHandle` JSDoc in `authConfig.ts` to state the return value must match `HANDLE_REGEX`.

### Smaller notes

- Line 78 changed `validateHandle('user123')` → `validateHandle('user-123')` rather than adding a case. Purely alphanumeric handles are the most common form and now have no direct coverage — worth keeping both.
- `/[^a-zA-Z0-9_-]/g` in `utils.ts` restates `HANDLE_REGEX`'s character class inline. If the rule changes these will drift; consider deriving both from one constant.
- The regex permits all-punctuation handles (`___`, `-_-`), and the sanitizer actively produces `___` for non-ASCII local parts. Not blocking, but requiring at least one alphanumeric character would be a reasonable tightening.

ankit-0903:
@omegascorp Thanks for the thorough review! I've updated the PR with both
requested changes:

   - *Legacy Handles:* In handleUpdateProfile, if props.handle matches
   profile.handle (unchanged), handle is dropped prior to
   validateProfileFields. Legacy users with pre-existing handles (e.g.
   john.doe) can now update their firstName/lastName without being blocked,
   while changes to new handles are still strictly validated. Added a unit
   test covering this behavior.


   - *JSDoc Update:* Updated generateHandle JSDoc in authConfig.ts to
   document that returned handles must match HANDLE_REGEX.


   - *Tests & Coverage:* Preserved test coverage for both user123 and
   user-123 in validators.test.ts.

All unit tests, linting, and formatting checks are passing!

omegascorp (approved):
Thank you, looks good

ankit-0903:
Thank you @omegascorp! Glad to see PR #335 merged as well!

PR #342, opened 2026-08-11
Claude Code plugin and browser-based setup

artahian:
Adds the Modelence plugin for Claude Code (output style, skills, MCP server) and the new `npx modelence setup` flow that authorizes in the browser instead of pasting a token from the dashboard.

Don't merge before the Studio PR (modelence/studio#428) is deployed — the new setup flow depends on its auth endpoints. CI will stay red until the version is bumped from the dev tag to a plain 0.23.0 release, which is intentional.

---

> [!NOTE]
> **Medium Risk**
> Changes cloud connect/setup auth, environment leasing, and first-run dev behavior; depends on Studio deploy for the new setup endpoints.
> 
> **Overview**
> Ships a **Claude Code plugin** (marketplace entry, framework output style, `modelence-patterns` / `expo-router` skills, docs + cloud MCP with `.modelence.env` token auth) and documents it in a new **AI Coding Agents** guide. **`npx modelence setup`** can run without a dashboard token: browser device auth picks the environment, writes `.modelence.env`, records **`appId`** in commit-friendly `.modelence/project.json`, and optionally seeds **`.claude/settings.json`** for the plugin.
> 
> **Local dev / cloud attachment** gets stable **instance IDs** (dev uses instance id as container id), **`--takeover`** on `modelence dev` to force-connect when another holder has the lease (409 guidance), **`MODELENCE_RUNTIME`**, default **`_system.site.url`** to the local server via `getLocalSiteUrl`, and **exit on detach** when sync reports another instance took over.
> 
> **Unconnected clones** see a built-in **setup screen** in dev (server `isSetupRequired`, client polls `_system.setupStatus` and reloads after setup). CLI auth no longer persists tokens to disk. Package bumps to **0.23.0** with **Node ≥22**; CI blocks merging while `modelence` version isn’t plain semver.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02e7c98ac06f3786d56a9890e7d106ce7fde8333. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/modelence/src/bin/setup.ts
```diff
@@ -2,22 +2,39 @@ import { promises as fs } from 'fs';
 import { join } from 'path';
 import { parse as parseEnv } from 'dotenv';
 import { createInterface } from 'readline';
+import { authenticateCli } from './auth';
 
 const MODELENCE_ENV_FILE = '.modelence.env';
+const MODELENCE_DIR = '.modelence';
+const PROJECT_FILE = 'project.json';
```
omegascorp, line 9: .modelence/project.json is gitignored in the template; should it be committed?

packages/modelence/src/bin/auth.ts
```diff
@@ -14,16 +29,19 @@ export async function authenticateCli(host: string) {
   }
 
   const { code, verificationUrl } = await response.json();
+  const hintParams = new URLSearchParams({
+    ...(pickEnvironment ? { pick: 'environment' } : {}),
+    ...(appId ? { appId } : {}),
+  });
+  const url = hintParams.size > 0 ? `${verificationUrl}&${hintParams}` : verificationUrl;
```
omegascorp: This feature requires Node 19 or higher, but we didn't declare a Node.js version requirement, I think we should start requiring node >= 22 to avoid issues with the node version

packages/modelence/src/client/AppProvider.tsx
```diff
@@ -31,11 +41,14 @@ function hasServerRenderedMarkup(): boolean {
...
 
-export function AppProvider({ children, loadingElement }: AppProviderProps) {
+export function AppProvider({ children, loadingElement, setupElement }: AppProviderProps) {
   const isServer = typeof window === 'undefined';
   const [isLoading, setIsLoading] = useState(
     () => !isServer && !isSessionInitialized() && !hasServerRenderedMarkup()
   );
+  const [needsSetup, setNeedsSetup] = useState(
+    () => !isServer && isSessionInitialized() && _isSetupRequired()
+  );
```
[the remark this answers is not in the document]
omegascorp: The same thing is pointed out by the local agent; here are additional details

[AppProvider.tsx](vscode-webview://09ield6oqm2ckod985abacamk2cj21n5ut1r0h57lt0svrs3p6mo/packages/modelence/src/client/AppProvider.tsx) sets initial state from isSessionInitialized() && _isSetupRequired(), then only updates it inside the initSession() branch:


if (isSessionInitialized()) {
  if (_isReconciliationPending()) {
    await reconcileSession();
  }
  return;          // <-- no setNeedsSetup
}
await initSession();
setNeedsSetup(_isSetupRequired());
The initial-state read runs during the first render, while hydrateSession() is called in [renderApp.tsx:93](vscode-webview://09ield6oqm2ckod985abacamk2cj21n5ut1r0h57lt0svrs3p6mo/packages/modelence/src/client/renderApp.tsx#L93) before the tree renders — so for the SSR path the lazy initializer does see the hydrated value and this works. But it works by ordering coincidence between two files, and the early-return branch has no setNeedsSetup, so any future change that defers hydration turns the setup screen off silently. A setNeedsSetup(_isSetupRequired()) before that return makes it robust for free.

Worth confirming SSR is even reachable in the setup-required state — if viteServer can't render without a backend, this is moot and the initializer could be simplified instead.

packages/modelence/src/client/AppProvider.tsx
```diff
@@ -54,15 +67,46 @@ export function AppProvider({ children, loadingElement }: AppProviderProps) {
...
+          {},
+          { errorHandler: () => {} }
+        );
+        if (!setupRequired) {
+          window.location.reload();
+        }
+      } catch {
+        // Dev server restarting — keep polling.
+      }
+    }, SETUP_POLL_INTERVAL);
```
omegascorp: setInterval might be throttled when it's in the background, because of throttling, multiple events can fire one after another. Can't we listen to an event via socketio instead?

omegascorp: Here is AI's feedback on this:

The setInterval callback is async and fires unconditionally every 3s regardless of whether the previous callMethod resolved. While the dev server is restarting, each request hangs until its TCP timeout, so requests accumulate. Low stakes on a dev-only screen, but a self-rescheduling setTimeout after each settle is the standard shape and is barely more code.

Also: window.location.reload() fires from inside the interval without clearing the timer first. Harmless in practice since the page unloads, but if the reload is blocked (rare) you get a reload loop.

artahian:
@artahian comment for myself:

Let's adjust `modelence setup` to automatically create `.claude/settings.json` with the modelence plugin enabled inside (if it doesn't exist already). If it exists but is disabled, it can also throw a warning.

packages/modelence/src/client/AppProvider.tsx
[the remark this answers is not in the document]
artahian: The true SSR mismatch here turns out to be unreachable for template apps — the template doesn't enable `ssr: true` — and cosmetic otherwise: it's dev-only, and React discards the server markup and client-renders, so the user still sees the setup screen (plus a console error). Also, the SSR renderer passes the session through context rather than hydrating the client session module, so the server-side initializer would read `false` even without the `isServer` guard. The ordering-coincidence point was fair though — added `setNeedsSetup(_isSetupRequired())` before the early return in 4f3d68b7 so the hydrated path no longer depends on `hydrateSession()` running before first render.

packages/modelence/src/bin/setup.ts
artahian: It's committed as of the template PR (modelence/app-builder-empty-project#28) — the gitignore there no longer ignores `.modelence` wholesale, only temp subpaths (`build/`, `cache/`, etc.), so `project.json` at the root gets committed as intended.

packages/modelence/src/bin/auth.ts
artahian: Agreed — added `"engines": { "node": ">=22" }` in 22039533.

packages/modelence/src/client/AppProvider.tsx
artahian: Switched to a self-rescheduling `setTimeout` in 4f3d68b7 — the next poll is only scheduled after the previous one settles (no pile-up during dev-server restarts, no burst after background-tab throttling), and after `window.location.reload()` it simply doesn't reschedule. Kept polling over a socket event: the setup screen exists precisely when the project has no backend, and a plain poll survives dev-server restarts with zero reconnect logic — 3s polling on a dev-only screen is cheap.

artahian:
^ Done in bf504ba4 — `modelence setup` now creates `.claude/settings.json` with the plugin enabled when the file doesn't exist (matching what the template ships). An existing settings.json is never modified: setup prints a hint when the plugin entry is missing, and a note when it's explicitly `false`, so a deliberate opt-out is never flipped back on.

packages/modelence/src/bin/setup.ts
```diff
@@ -50,6 +67,98 @@ function escapeEnvValue(value: string | number): string {
...
+  let content: string;
+  try {
+    content = await fs.readFile(settingsPath, 'utf8');
+  } catch {
+    // Missing — create it with just the plugin enabled.
+    try {
+      await fs.mkdir(join(process.cwd(), CLAUDE_DIR), { recursive: true });
+      await fs.writeFile(
+        settingsPath,
+        JSON.stringify({ enabledPlugins: { [CLAUDE_PLUGIN_ID]: true } }, null, 2) + '\n'
```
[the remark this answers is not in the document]
artahian: Good catch — confirmed against current Claude Code behavior: without `extraKnownMarketplaces`, the plugin reference can't be resolved unless the user has already added the marketplace manually. Fixed in 3b6ae886: the created settings.json now declares the marketplace alongside `enabledPlugins`, and the hint for existing settings files mentions it too. The template's committed `.claude/settings.json` (which actually had the same gap) and the docs example (which used the outdated array form of `enabledPlugins`) are updated as well.

omegascorp:
@artahian could you check the following issues:

1. SSR apps get a hydration mismatch on the setup screen
[ssr/render.tsx:85](/packages/modelence/src/ssr/render.tsx#L85) constructs <AppProvider loadingElement={loadingElement}> — setupElement is never threaded through, and on the server [AppProvider.tsx:48](/packages/modelence/src/client/AppProvider.tsx#L48) initializes needsSetup to false because isServer is true.

For an app with startApp({ ssr: true }) and no .modelence.env:

Server renders the real app tree. Every page component runs — the ones calling modelenceQuery against a nonexistent database included.
Client: hydrateSession sets setupRequired = true, the SSR marker is present so isLoading is false, isSessionInitialized() is true → first client render is SetupScreen.
hydrateRoot hydrates SetupScreen against server-rendered app markup → mismatch, full client re-render, React error in console.
The setup screen is precisely for people who can't run the app, so this is the state SSR users land in on a fresh clone. Two fixes needed: pass setupElement through SsrRenderOptions/renderSsrTreeStream, and let AppProvider decide needsSetup on the server too (the payload is already in scope via the SSR context — sessionPayload.setupRequired).

Related, smaller: renderApp reads setupElement from options and passes it to AppProvider, but _getSsrSnapshot() never forwards it, so even after fixing the server path you need the snapshot to carry it.

2. escapeEnvValue doesn't escape backslashes
[setup.ts:65-68](packages/modelence/src/bin/setup.ts#L65-L68):


return String(value).replace(/"/g, '\\"');
Values are written as KEY="…". dotenv unescapes \" inside double quotes, but a value containing a trailing backslash produces KEY="abc\" — the closing quote is consumed as an escape and the parse runs into the next line. MODELENCE_SERVICE_TOKEN is server-generated so this is unlikely to fire today, but MODELENCE_SERVICE_ENDPOINT is options.host, i.e. arbitrary user input from --host. Escape backslashes first:


return String(value).replace(/\\/g, '\\\\').replace(/"/g, '\\"');
Same file, related: [mcp-auth.js:33](/claude-plugins/modelence/scripts/mcp-auth.js#L33) reads the token with /^\s*MODELENCE_SERVICE_TOKEN\s*=\s*("?)(.*)\1\s*$/ and does not unescape \". So the two halves disagree about the escaping convention. Since the writer is the only producer of this file, aligning them is cheap — either both escape or neither does.

3. The verification URL assumes an existing query string
[auth.ts:36](/packages/modelence/src/bin/auth.ts#L36):


const url = hintParams.size > 0 ? `${verificationUrl}&${hintParams}` : verificationUrl;
& is hardcoded. This works only if Studio's verificationUrl already contains ?. It probably does today (it carries the code), but this is a silent-breakage coupling across two repos, and the fix is one line:


const url = new URL(verificationUrl);
if (pickEnvironment) url.searchParams.set('pick', 'environment');
if (appId) url.searchParams.set('appId', appId);
That also removes the hintParams.size branch entirely.

## Worth fixing
Polling errors now spam every setup run. [auth.ts:56-58](/packages/modelence/src/bin/auth.ts#L56-L58) — pollForToken throws on any non-2xx, and waitForAuth catches it with console.error('Error polling for CLI token:', error) every 5s. If Studio returns a non-200 for "not yet authorized" (a common device-flow convention), a user taking 60s to approve sees a dozen scary errors. This was tolerable when only deploy used it; setup is now the first command a new user runs. At minimum, don't log until several consecutive failures.

open() failure isn't handled. [auth.ts:43](/packages/modelence/src/bin/auth.ts#L43) — await open(url) rejects on headless/SSH/container. The URL and code are printed first, so the information is there, but the rejection propagates out of setup() and the user gets Setup failed: … instead of "open this URL manually." Wrap it in a try/catch.

.modelence/ is both committed state and build scratch. deploy writes .modelence/tmp/bundle.zip and .modelence/build/app.mjs into the same directory project.json now lives in. The comment claims "only designated subdirs like cache/ are temporary" — but tmp/ and build/ are too. If a project's .gitignore has .modelence/, project.json is silently never committed and the app-hint feature quietly does nothing. This depends on a gitignore change in the modelence/examples template repo, which isn't in this PR. Either flag it in the PR description as a cross-repo dependency, or make setup print a hint when the file it just wrote is git-ignored.

Docs claim something not yet true. [ai-coding-agents.mdx](/docs/web/ai-coding-agents.mdx) says "New Modelence projects come with the plugin already wired up" and "this is the same file new projects ship with." Templates live in modelence/examples, not this repo. Ship the template change first or soften the wording.

## Simplifications
Two stacked comments on one function. [setup.ts:70-88](/packages/modelence/src/bin/setup.ts#L70-L88) — a long block explaining recordProjectAppId sits directly above readProjectAppId, followed by a second block that actually describes readProjectAppId. The first is orphaned; move it onto recordProjectAppId where it belongs.

The errorHandler: () => {} + catch {} pair is redundant. [AppProvider.tsx:88-92](/packages/modelence/src/client/AppProvider.tsx#L88-L92) — callMethod calls the handler and rethrows, so the catch is doing all the work. Drop the errorHandler option; the catch alone is enough and reads clearer.

reconcileSession doesn't refresh setupRequired. [session.ts:130](/packages/modelence/src/client/session.ts#L130) destructures { configs, session, user } and drops setupRequired, then AppProvider re-reads _isSetupRequired() right after and gets the stale hydrate value. Harmless today (the flag can't change without a server restart), but the re-read at [AppProvider.tsx:69](/packages/modelence/src/client/AppProvider.tsx#L69) is written as if it might have changed. Either set it in reconcileSession or drop the re-read.

artahian:
Went through all of these — thanks for the depth. Fixed in 46a51533:

- **Verification URL**: now built with `new URL()` + `searchParams`, no more coupling to Studio's query string.
- **`open()` failure**: wrapped — headless/SSH prints "open the URL manually" instead of failing setup.
- **Stacked comments**: moved onto their functions, and the comment now lists the actual temp subdirs (`build/`, `cache/`, `tmp/`).

On the rest:

**1 — SSR setup screen**: agreed it's real for `ssr: true` apps, and the "page components run against a nonexistent DB during SSR" point is fair — that part I underweighted in my earlier reply. It's dev-only and no template enables SSR, so deferring out of this PR: the fix needs `setupElement` threaded through `SsrRenderOptions`/`viteServer` plus a server-side `setupRequired` path in `AppProvider` (the SSR session resolver doesn't expose it today). One correction: `_getSsrSnapshot()` does carry `setupElement` — the snapshot is the options object verbatim; the drop is in `viteServer.ts` which only forwards four fields.

**2 — escapeEnvValue**: tested against the dotenv we actually ship (16.5.0) — the trailing-backslash failure doesn't fire: `A="trailing\"` parses to `trailing\` and the next line is unaffected, because dotenv's `[^"]` alternation backtracks over the backslash. Escaping backslashes would make things worse, not better: dotenv v16 doesn't unescape `\\` (or `\"` — the existing quote-escaping is already lossy on round-trip). So if we align writer and readers, it's by removing escaping, not adding it. Both values are a server-generated token and a URL, so leaving as-is for now.

**Polling errors**: Studio returns 200 `{status: "pending"}` while unapproved, so `pollForToken` doesn't throw during a normal wait — the console.error only fires on genuine network/server failures, which is worth keeping visible.

**errorHandler noop**: it's load-bearing, not redundant — `callMethod` falls back to the global handler when no per-call handler is given, then rethrows. Dropping the noop would fire the app's global error handler (console/toast) every 3s while the dev server restarts. The noop suppresses the default handler; the catch handles the rethrow.

**reconcileSession / setupRequired**: correct that it's dropped, and harmless as you said — both payloads come from the same server boot. The re-read in AppProvider is the ordering fix from your earlier comment, so keeping both as-is rather than undoing one to justify the other.

**.modelence layout**: intentional — the `.modelence/` root is committed CLI-managed state, temp subdirs are gitignored. That's exactly what modelence/app-builder-empty-project#28 ships. The examples repo gets the same treatment (plus `.claude/settings.json` and — noticed while checking — a missing `.modelence.env` gitignore entry in some templates) right after this PR merges: it can't happen earlier because create-modelence-app serves examples `main` directly and the marketplace file only exists on this branch until merge. Same reason the docs wording stays: it becomes true with that examples update.

artahian:
Pushed a few more things since the last round — all of it the framework half of the attach/detach switch (MOD-578), paired with modelence/studio#451.

- **Instance identity.** Dev runtimes no longer use the `MODELENCE_CONTAINER_ID` that setup wrote — every machine got the same one. The container ID is now a hash of a random per-machine ID (`~/.modelence/machine-id`, kept outside the repo so it can't be committed and cloned), the working directory, and the environment ID. The same directory reconnecting presents the same ID; the same directory pointed at another environment gets a new one. Subsumes MOD-514.
- **`MODELENCE_RUNTIME`.** `modelence setup` writes `local`; Studio sandboxes pre-set `sandbox` via the task definition. Classification only — it decides whether Studio can stop an instance itself or merely signal it.
- **Takeover.** Connect now reports `runtime`, and `modelence dev --takeover` adds `force`. A 409 without the flag prints who holds the environment and how to take it over, then exits — no stack dump for what's an expected condition on a dev machine. Only the `local` runtime may force; Studio enforces the same rule server-side.
- **Kill switch.** `/api/sync` can answer `detached`, and a dev runtime prints the reason and exits. That's what makes the dashboard's Disconnect and a takeover actually stop the other side. Guarded by the dev-runtime check so a Studio bug can never take down cloud containers.
- **Local site URL.** With `MODELENCE_RUNTIME=local`, `_system.site.url` now defaults to `http://localhost:<port>` instead of the environment URL from the cloud config, so OAuth callbacks and email links hit the server that's actually running. An explicit `MODELENCE_SITE_URL` still wins.

Published as `0.23.0-setup.dev.9`.

## Testing

Needs modelence/studio#451 deployed for the Studio side (dev is applied already).

1. `npm i modelence@0.23.0-setup.dev.9` in a test project, then `npx modelence setup` — browser device auth, pick the environment.
2. `npm run dev` — attaches, stops the cloud sandbox, and the dashboard's toolbar switches to a green "Connected locally" with your hostname.
3. Run `npm run dev` from a second checkout of the same environment: refused, naming the holder. `npm run dev -- --takeover` takes it, and the first one exits on its next heartbeat.
4. Dashboard → **Connected locally** → **Disconnect**: the local dev server exits within a few seconds and the sandbox is free to wake again.
5. Hit the environment URL while local holds it — the wake page says it's running locally and won't start the sandbox. Stop the local server and the button works again (the auto-start stays suppressed until you click it).

Pre-0.2.0 plugin installs need `claude plugin update` to pick up the new plugin.

### Pointing the plugin's MCP at a local Studio

`MODELENCE_MCP_URL` overrides the plugin's cloud MCP URL, but it has to be in the **process environment when Claude Code starts** — the `env` block in `.claude/settings.json` is not used for MCP URL expansion (it was until extension 2.1.220, which is why this used to work). For the VSCode extension, add to your user `settings.json`:

```json
"claudeCode.environmentVariables": [
  { "name": "MODELENCE_MCP_URL", "value": "http://localhost:3000/mcp" }
]
```

and reload the window. For a terminal `claude`, just export it in the shell before launching. The service token from `.modelence.env` is attached automatically for localhost and for the prod default, and withheld from anything else.

One gotcha that cost me a while: after a failed connect, Claude Code caches the server in `~/.claude/mcp-needs-auth-cache.json` and new sessions skip connecting entirely — still showing "Needs Auth" even once the URL is right. Delete the `plugin:modelence:modelence` entry (or re-auth via `/mcp`) to make it retry.

omegascorp (approved):
Code looks good, let's ship it. One thing I didn't get: how do we install the claude plugin?

omegascorp:
@artahian I found the answer to my question in the docs

PR #349, opened 2026-08-23
fix(auth): prevent signup email enumeration

### LINKED ISSUE #339: Signup endpoint leaks whether an email is registered (email enumeration)
Follow-up from review on PR #305 (https://github.com/modelence/modelence/pull/305).

`signup.ts` throws `User with email already exists: ${existingEmail?.address}` (lines 55 and 137) — a direct, unauthenticated confirmation that an email is registered. Login was hardened against timing-based enumeration, but an attacker can just hit signup for an instant answer.

`resetPassword.ts` already does this correctly and returns a generic success. Consider returning a generic message from signup too (e.g. send a verification email regardless, like reset password does).

mikemikimike:
Closes #339.

## Summary

The password signup endpoint included the submitted address or account status in duplicate-email errors, allowing unauthenticated callers to confirm whether an email was registered.

## Changes

- Return the fixed Unable to create account message for active duplicate accounts, disabled accounts, and concurrent unique-index failures.
- Preserve signup error callbacks while removing account-state details from the client-visible error.
- Add regression coverage for existing emails, case variants, disabled accounts, and concurrent duplicate inserts.

## Compatibility

Duplicate-email signups still fail and invoke the same error callbacks; only the client-visible error text changes. The email address and disabled status are no longer disclosed.

## Verification

- 
pm test -- --run src/auth/signup.test.ts — 19 passed.
- 
pm test -- --run src/auth — 296 passed across 22 files.
- 
pm run lint:check — passed.
- 
pm run build — passed (ESM and DTS).

The test runner emitted unrelated tsconfig discovery warnings from other directories in the shared workspace, but the targeted commands exited successfully.

---

> [!NOTE]
> **Low Risk**
> Security-hardening change limited to client-visible error text; signup failure behavior and callbacks are unchanged.
> 
> **Overview**
> Password signup no longer leaks whether an email is registered or whether the account is disabled. **Duplicate-email**, **disabled-account**, and **concurrent unique-index** failures all return the same client message: `Unable to create account`, instead of messages that included the address or deletion/support wording.
> 
> Signup still fails on those paths and **`onSignupError` / `signup.onError`** still run with the generic error. Tests were updated to assert the new message (including case-insensitive duplicates) and to cover disabled accounts without exposing status.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00ad426c9c3eda29f5e984f40953ca35f0c9ccd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/signup.ts
+++ b/packages/modelence/src/auth/signup.ts
@@ -47,13 +47,12 @@ export async function handleSignupWithPassword(
     );
 
     if (existingUser) {
-      const existingEmail = existingUser.emails?.find((e) => e.address.toLowerCase() === email);
       if (existingUser.status === 'disabled') {
         throw new Error(
           `User is marked for deletion, please contact support if you want to restore the account.`
         );
       }
-      throw new Error(`User with email already exists: ${existingEmail?.address}`);
+      throw new Error('Unable to create account');
     }
 
     await authConfig.onBeforeSignup?.({
@@ -143,7 +142,7 @@ export async function handleSignupWithPassword(
         // A concurrent signup won the race between the findOne check above and
         // this insert; the unique emails.address index rejected the duplicate.
         // Surface the same message the pre-check would have.
-        throw new Error(`User with email already exists: ${email}`);
+        throw new Error('Unable to create account');
       }
       throw error;
     }
--- a/packages/modelence/src/auth/signup.ts
+++ b/packages/modelence/src/auth/signup.ts
@@ -47,9 +47,6 @@ export async function handleSignupWithPassword(
     );
 
     if (existingUser) {
-      if (existingUser.status === 'disabled') {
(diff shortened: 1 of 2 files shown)


packages/modelence/src/auth/signup.ts
```diff
@@ -47,13 +47,10 @@ export async function handleSignupWithPassword(
     );
 
     if (existingUser) {
-      const existingEmail = existingUser.emails?.find((e) => e.address.toLowerCase() === email);
       if (existingUser.status === 'disabled') {
```
omegascorp: Please remove this condition, both messages are identical

omegascorp (approved):
Thank you, looks good

PR #352, opened 2026-08-28
feat: allow to configure trusted proxies

omegascorp:
## Summary

Adds secure, configurable trusted-proxy handling for client IP resolution and IP-based rate limiting.

Previously, Modelence selected the leftmost `X-Forwarded-For` value directly. Because clients can supply this header, an attacker could rotate its value to bypass authentication rate limits.

## Changes

- Use Express trusted-proxy chain resolution through `req.ip`.
- Add `security.trustedProxies` to `startApp()`.
- Add native `MODELENCE_TRUSTED_PROXIES` environment-variable support.
- Give the environment variable precedence over `startApp()` configuration.
- Preserve the legacy trust-all behavior when neither setting is provided to avoid breaking existing reverse-proxy deployments.
- Add documentation and regression tests.

### Environment configuration

```env
MODELENCE_TRUSTED_PROXIES="loopback,linklocal,uniquelocal,10.0.0.0/8"
```

### Application configuration

```ts
startApp({
  security: {
    trustedProxies: ['loopback', 'linklocal', 'uniquelocal'],
  },
});
```

## Backward compatibility

When no trusted-proxy configuration is provided, Modelence continues using the historical trust-all behavior. This prevents users behind an existing load balancer or router from unexpectedly sharing one rate-limit identity.

The legacy default remains susceptible to spoofed `X-Forwarded-For` values. Hosted environments should configure `MODELENCE_TRUSTED_PROXIES` before upgrading to activate secure proxy-chain resolution without disrupting client IP handling.

---

> [!NOTE]
> **High Risk**
> Changes how client IP and public base URL are derived behind proxies, directly affecting authentication rate limits and callback URLs; misconfiguration can break limits or leave spoofing possible until explicit trusted proxies are set.
> 
> **Overview**
> Fixes **IP-based rate limit bypass** by no longer trusting the leftmost `X-Forwarded-For` value. Client IP for `connectionInfo.ip` now follows Express **`trust proxy`** resolution via `req.ip`, and **`baseUrl`** uses `req.protocol` so untrusted callers cannot spoof HTTPS for auth callbacks.
> 
> Adds **`security.trustedProxies`** on `startApp()` and **`MODELENCE_TRUSTED_PROXIES`** (env wins over app config). When neither is set, **trust-all** behavior remains for upgrade compatibility. Optional **`clientIpHeader`** (e.g. `cf-connecting-ip`) is honored only when the socket peer is a trusted proxy.
> 
> Documentation covers reverse-proxy and Cloudflare setup; tests cover proxy config, IP resolution, and env parsing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4906867982a301676c50cd264a4e18c298d9d547. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/web/core-concepts/environment-and-setup.mdx
+++ b/docs/web/core-concepts/environment-and-setup.mdx
@@ -66,6 +66,7 @@ Every `MODELENCE_*` variable (plus the few non-prefixed ones) that the framework
 | `MODELENCE_SITE_URL` | string | — | `packages/modelence` (public site URL, used by auth callbacks and emails) |
 | `PORT` | number | `3000` | `packages/modelence/app/server.ts` |
 | `MODELENCE_PORT` | number | falls back to `PORT` | `packages/modelence/app/server.ts` |
+| `MODELENCE_TRUSTED_PROXIES` | string (comma-separated addresses/ranges) | all proxies trusted for backward compatibility | `packages/modelence/app/server.ts` |
 | `NODE_ENV` | `development` \| `production` | `development` | Build pipeline, cookie `secure` flag, Vite dev server |
 | `MODELENCE_LOG_LEVEL` | `debug` \| `info` \| `warn` \| `error` | `info` | `packages/modelence/telemetry` |
 | `MODELENCE_ENV_TYPE` | string | — | `packages/modelence/config` (`_system.env.type`) |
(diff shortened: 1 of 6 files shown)


artahian (commented):
@omegascorp overall looks ok to me, but I want to make sure that this matches the Cloudflare recommendations (for ourselves and anyone else running behind a Cloudflare proxy).

Can you please check if what we're doing here is in line with the below? There is a specific `cf-connecting-ip` for Cloudflare's case:

https://developers.cloudflare.com/fundamentals/reference/http-headers/

omegascorp:
@artahian done, added support of `CF-Connecting-IP`

artahian (approved):

Issue #353, opened 2026-08-30
Security: TOCTOU race condition in consumeRateLimit allows rate-limit bypass during concurrent bursts


ankit-0903:
### Description
In `packages/modelence/src/rate-limit/rules.ts`, the `checkRateLimitRule` function reads the current window count via `dbRateLimits.findOne(...)`, checks if `count >= rule.limit`, and subsequently updates the record using `dbRateLimits.upsertOne(...)`.

Because the read (`findOne`) and write (`upsertOne`) operations are separate async steps rather than an atomic update, concurrent HTTP requests during burst traffic (e.g. rapid credential stuffing, signup spam, or parallel API calls) execute `findOne` simultaneously. All concurrent requests read a stale count below `rule.limit` before any request increments the counter, bypassing rate limits entirely.

### Actual vs Expected
- **Actual**: Concurrent requests within the same millisecond window read stale count data and bypass configured rate limits.
- **Expected**: Rate limit checking and incrementing should be atomic so that concurrent requests strictly respect `rule.limit`.

### Affected Code
- **File**: `packages/modelence/src/rate-limit/rules.ts` (`checkRateLimitRule`)

PR #355, opened 2026-08-30
refactor(rate-limit): extract shared filter in checkRateLimitRule

ankit-0903:
Added test suite coverage in `rules.test.ts` verifying rate-limit threshold enforcement when count threshold is reached. All unit tests, formatting, and linting checks are passing cleanly.

---

> [!NOTE]
> **Medium Risk**
> Touches core rate-limit consumption logic used for abuse protection; the change is small but incorrect behavior could allow extra requests or block legitimate ones.
> 
> **Overview**
> **Fixes a TOCTOU-style consistency issue** in `checkRateLimitRule` by building one `filter` object and passing it to both `findOne` and `upsertOne`, so the read and increment always target the same rate-limit document instead of duplicating filter literals.
> 
> Adds a unit test where the stored count already equals the rule limit (`windowCount: 3`, `limit: 3`), asserting `consumeRateLimit` throws and **does not** call `upsertOne`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6869abfa475012e2c794639b82ae87f98f255954. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/rate-limit/rules.test.ts
+++ b/packages/modelence/src/rate-limit/rules.test.ts
@@ -130,4 +130,28 @@ describe('rate-limit/rules', () => {
 
     expect(mocks.upsertOne).not.toHaveBeenCalled();
   });
+
+  test('consumeRateLimit throws when rate limit count meets or exceeds limit threshold', async () => {
+    vi.useFakeTimers().setSystemTime(new Date('2024-01-01T00:01:00.000Z'));
+    const { initRateLimits, consumeRateLimit, mocks } = await loadModule();
+
+    initRateLimits([{ bucket: 'login', type: 'ip', window: 60_000, limit: 3 }]);
+
+    mocks.findOne.mockResolvedValue({
+      bucket: 'login',
+      type: 'ip',
+      value: '10.0.0.1',
+      windowMs: 60_000,
+      windowStart: new Date('2024-01-01T00:01:00.000Z'),
+      windowCount: 3,
+      prevWindowCount: 0,
+      expiresAt: new Date('2024-01-01T00:03:00.000Z'),
+    } as never);
+
+    await expect(
+      consumeRateLimit({ bucket: 'login', type: 'ip', value: '10.0.0.1' })
+    ).rejects.toThrow('Rate limit exceeded for login');
+
+    expect(mocks.upsertOne).not.toHaveBeenCalled();
+  });
 });
--- a/packages/modelence/src/rate-limit/rules.ts
+++ b/packages/modelence/src/rate-limit/rules.ts
@@ -50,12 +50,14 @@ async function checkRateLimitRule(rule: RateLimitRule, value: string, createErro
       : new RateLimitError(`Rate limit exceeded for ${rule.bucket}`);
   };
 
-  const record = await dbRateLimits.findOne({
+  const filter = {
     bucket: rule.bucket,
     type: rule.type,
     value,
     windowMs: rule.window,
-  });
+  };
+
+  const record = await dbRateLimits.findOne(filter);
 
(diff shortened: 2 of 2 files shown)


omegascorp (commented):
@ankit-0903, the PR claims to fix a TOCTOU race condition in consumeRateLimit, but changes only the test

ankit-0903:
@omegascorp I've updated the branch and force-pushed the implementation
changes in packages/modelence/src/rate-limit/rules.ts alongside the unit
tests. All 4 CI checks are passing!

omegascorp (changes requested):
Thanks for digging into this, and for opening #353 — the underlying issue is real and worth fixing. But I don't think this PR can be merged as-is, because it doesn't fix the race it claims to.

### The change itself is fine, just not a fix

Extracting the duplicated filter literal into a `filter` const is a reasonable readability improvement — it stops the two filters drifting apart in future edits. No objection to that on its own.

The problem is that both filter literals were already byte-for-byte identical, so hoisting them into a variable is a pure refactor with zero behavioral change.

### The TOCTOU race is still fully present

The race described in #353 is that `findOne` (read) and `upsertOne` (write) are two separate round-trips to MongoDB. N concurrent requests all complete their `findOne` before any of them reaches `upsertOne`, so all N read the same stale count and all N pass the `count >= rule.limit` check:

```ts
const record = await dbRateLimits.findOne(filter);  // request A, B, C all read here
// ...
if (count >= rule.limit) {                          // ...all three see the same stale count
  throw createRateLimitError();
}
// ...
await dbRateLimits.upsertOne(filter, modifier);     // only now does anyone increment
```

Sharing a filter variable between those two calls doesn't change the interleaving — it's the same two non-atomic operations in the same order. The bypass remains exploitable exactly as described in the issue.

A real fix needs the read and increment collapsed into one atomic operation, e.g. `findOneAndUpdate` with `$inc` and `returnDocument: 'after'`, then throwing based on the returned count. That's a genuine design change (including deciding whether to decrement on rejection or accept a slight overcount), not present here.

### The new test doesn't cover the race either

The added test is close to a duplicate of the existing `'consumeRateLimit throws when limit exceeded in current window'` case — same single-threaded shape, same assertions, just `windowCount: 3, limit: 3` instead of `1/1`. Exercising the `>=` boundary at equality has some marginal value, but it makes one sequential call, so it would pass unchanged whether or not the TOCTOU bug exists. A test for this needs two interleaved in-flight calls.

### Suggested path forward

Either one works for me:

1. **Retitle to something like `refactor(rate-limit): extract shared filter`, and drop the `Fixes #353`.** Happy to merge on that basis — it leaves #353 open for a proper fix.
2. **Implement the atomic `findOneAndUpdate` and add a test with genuinely concurrent calls.** More work, but it actually closes the issue.

My only real concern with merging as-is is that it would auto-close a live security issue while the vulnerability is still there.

One more note: the Cursor Bugbot summary in the description says this fixes "a TOCTOU-style consistency issue... so the read and increment always target the same rate-limit document." That's misleading — the filters were already identical and always targeted the same document. Worth ignoring that framing.

ankit-0903:
@omegascorp Thanks for the detailed feedback! That makes total sense. I've updated the PR title to `refactor(rate-limit): extract shared filter in checkRateLimitRule` and removed `Fixes #353` from the description so #353 remains open for a dedicated atomic fix.

omegascorp (approved):

omegascorp:
@ankit-0903 thank you, merged

ankit-0903:
Thank you @omegascorp! Glad to contribute.

PR #359, opened 2026-08-31
feat: cors allowed origins (MOD-599)

### LINKED ISSUE #358: Add CORS support (security.allowedOrigins) — Expo Web cannot call the API
## Problem

The Modelence server sends no CORS headers, so any browser client on a different origin is blocked from reading API responses. The concrete case is **Expo Web**: Metro serves the app on `:8081` while the API runs on `:3000`, and different ports are different origins, so every method call fails.

The browser console shows:

```
Access to fetch at 'http://localhost:3000/api/_internal/method/_system.session.init'
from origin 'http://localhost:8081' has been blocked by CORS policy:
No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

The request itself **succeeds** — the server returns `200` — and the browser discards the response because it carries no CORS headers. `fetch` then rejects with a bare `TypeError: Failed to fetch`, which is indistinguishable from an unreachable server. Only the browser console tells them apart, and `curl` reports a clean `200` because it does not enforce CORS. This makes the failure very easy to misdiagnose as a network or config problem.

## Why it is not covered today

- `SecurityConfig` accepts only `frameAncestors`, which emits CSP `frame-ancestors`. That controls **iframe embedding**, not whether JS on another origin may read a fetch response. The two are unrelated mechanisms.
- There is no CORS middleware anywhere in the Express app, and no `cors` dependency. `securityHeadersMiddleware()` sets only `Content-Security-Policy` and `X-Frame-Options`.

Verified against a running dev server — `200 OK` with no `Access-Control-*` headers on the response, even with an `Origin` request header present.

## Impact

Every Modelence project with an Expo mobile app hits this the first time it opens the web preview locally. Native iOS/Android are unaffected (no CORS enforcement), so the workaround today is "test on a simulator or device", which is a poor first-run experience.

Studio-hosted sandboxes do not hit this because the sandbox router adds `Access-Control-Allow-Origin: *` to all proxied traffic — see the constraint below.

## Proposal

Add an opt-in `allowedOrigins` to the existing `SecurityConfig`:

```ts
startApp({
  modules: [...],
  security: {
    allowedOrigins: ['http://localhost:8081'],
  },
});
```

When set, a middleware echoes the request `Origin` if it is in the list, alongside `Access-Control-Allow-Credentials`, `Access-Control-Allow-Headers`, `Access-Control-Allow-Methods` and `Vary: Origin`, and answers `OPTIONS` preflights directly.

Design notes:

- **Echo the matched origin, never `*`.** A wildcard is rejected by browsers on credentialed requests, and web clients use cookie-based flows. `Vary: Origin` is required so a cache cannot serve one origin's response to another.
- **Exact-match origins only.** CORS headers take one concrete origin, so patterns like `http://localhost:*` (valid in CSP) cannot be expressed here.
- **Opt-in only — no implicit dev default.** See the constraint below.

## Constraint: do not collide with the sandbox router

Studio sandboxes sit behind a router that already sets `Access-Control-Allow-Origin: *` on **all** proxied tenant traffic, before the proxy to the app. If the SDK also emitted a header there, responses would carry two `Access-Control-Allow-Origin` values, which browsers treat as invalid — breaking every studio mobile preview.

Keeping this strictly opt-in avoids the collision by construction: sandboxes never set `allowedOrigins`, so the SDK stays silent and the router's header remains the only one. This is why no `NODE_ENV`-based automatic localhost allowance is proposed — sandboxes run with `NODE_ENV=development`, so that heuristic would fire exactly where it must not.

## Follow-up (separate, not in scope here)

A network-level `fetch` rejection is always a connectivity/CORS failure and never an application error, but it currently surfaces as an unadorned `TypeError: Failed to fetch` from a minified chunk. Wrapping it with the attempted base URL and a pointer to the console would make this class of problem self-diagnosing.

omegascorp:
* Allow configuring CORS

Closes: #358 

---

> [!NOTE]
> **Medium Risk**
> Changes cross-origin response headers and preflight behavior on API/auth routes; misconfiguration could block clients or widen API access, though SSR remains excluded and the feature is opt-in.
> 
> **Overview**
> Adds **opt-in CORS** via `security.allowedOrigins` on `startApp`, aimed at cross-origin browser API clients (e.g. Expo Web on another port).
> 
> **Configuration:** New `allowedOrigins` on `SecurityConfig` is validated and normalized at startup (exact `scheme://host[:port]`, no wildcards/paths/credentials); invalid entries throw instead of failing silently in the browser. When unset, **no** CORS headers are sent (avoids duplicating proxy CORS).
> 
> **Server wiring:** Global middleware only handles **OPTIONS preflights** (204/403, `Access-Control-Max-Age`). Response CORS headers are attached **per route** on module routes (any path), `/api/_internal/auth/*`, and the internal method endpoint—not on the SSR `*` catch-all. Allowed origins get echoed `Access-Control-Allow-Origin`, credentials, methods, reflected preflight headers, `Vary: Origin`, and `Access-Control-Expose-Headers: X-Modelence-Error-Code`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16c2f50bdb1d9b67fcafe32ca54d38f75bd2afac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/modelence/src/app/server.ts
```diff
@@ -348,6 +349,57 @@ function securityHeadersMiddleware(): express.RequestHandler {
...
+      // this the error-code header is hidden cross-origin, so MethodError.code
+      // silently becomes undefined and clients cannot branch on the error kind.
+      res.setHeader('Access-Control-Expose-Headers', 'X-Modelence-Error-Code');
+      // The response body varies by Origin, so a shared cache must not reuse
+      // one origin's response for another.
+      res.setHeader('Vary', 'Origin');
+    }
+
+    // Answer preflights here — they must never reach route handlers.
+    if (req.method === 'OPTIONS') {
```
eduard-piliposyan: This part runs before next(), and the middleware is registered before `registerModuleRoutes`. `HttpMethod `includes 'options' (and 'all' / 'use'), so any user-defined OPTIONS route becomes unreachable the moment CORS is enabled, and a plain OPTIONS /health from a non-browser client gets a 403. A preflight is specifically OPTIONS plus an Access-Control-Request-Method header. So let's handle that case with something like this. 
```
const isPreflight = req.method === 'OPTIONS' && typeof req.headers['access-control-request-method'] === 'string';
if (!isPreflight) { next(); return; }
res.sendStatus(isAllowed ? 204 : 403);
```

eduard-piliposyan (commented):
Here are some more issues, maybe not all of them are worth to handle, so please check @omegascorp 
Alo does the websocket layer do its own Origin check? If so, it probably needs to read the same allowlist, otherwise enabling allowedOrigins fixes method calls but leaves websockets failing.
- [ ] **`Vary: Origin` is only set on the match path, and `setHeader` clobbers any existing `Vary`.** When the origin isn't allowed (or there's no `Origin`), no `Vary` is sent, so a CDN or shared proxy can cache that header-less response and later serve it to an allowed origin. Set it whenever the middleware is active, not only on a match, and use `res.vary('Origin')` (which appends) instead of `res.setHeader`. Preflight responses should also vary on `Access-Control-Request-Headers`, since we reflect it back.
- [ ] **No `Access-Control-Max-Age`, so every cross-origin method call is two round trips.** The client always sends `Content-Type: application/json`, so every call is preflighted, and without `Max-Age` the browser re-preflights each time. Setting `Access-Control-Max-Age: 600` on the 204 halves the request count for the Expo Web flow.
- [ ] **Configured origins are matched by exact string with no normalization or validation.** `http://localhost:8081/` with a trailing slash, `http://LOCALHOST:8081`, or a full URL with a path silently never matches, and the developer just sees a generic CORS error with no hint why. Normalize entries (trim, lowercase, strip trailing slash) and validate at startup, throwing a clear error for anything that isn't a `scheme://host[:port]` origin, including `'*'`.
- [ ] **CORS is applied to every route, not just the API.** The middleware sits at the top of the stack, so allowlisted origins can read SSR pages and static responses too, with credentials. It's an explicit allowlist so this isn't a hole, but the config name and docs read as "API access". Scoping the middleware to `/api/` would match the stated intent.
- [ ] **The JSDoc claim about credentialed requests is only true same-site.** The `authToken` cookie is `sameSite: 'lax'`, so a genuinely cross-site setup (`app.example.com` calling `api.other.com`) still won't send the cookie regardless of `Access-Control-Allow-Credentials`. It works for the Expo case because `localhost:8081` and `localhost:3000` are the same site with different ports. Worth one line in the JSDoc.

artahian (commented):
@omegascorp I'm not sure I get it - how did this work with the old Expo web URL which was a 3rd party one?

Also, do we have any special routes that should be controlled separately? For mobile previews specifically, what happens if a user is running it on a different port? Will allowedOrigins support a general localhost wildcard with all ports allowed?

omegascorp:
@eduard-piliposyan all updated, please check again

omegascorp:
@eduard-piliposyan reminder to check the PR, there is a small conflict but it can be resolved before merging

eduard-piliposyan (approved):
There are some merge issues, the rest looks good!

PR #361, opened 2026-09-02
fix: improve oauth error message

omegascorp:
I've faced this case multiple times, and it was never a possible CSRF attack; my sessions just expired. I've added oauthErrorRedirectUrl and deprecated errorComponent, I think it will do the job better. errorComponent requires an HTML page, which is not a good fit for the framework.

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth error delivery and redirect behavior on provider callbacks; changes are opt-in via config and preserve JSON/errorComponent fallbacks, but misconfigured redirect URLs could still strand users until fallback.
> 
> **Overview**
> Adds **`auth.oauthErrorRedirectUrl`** so failed **web** OAuth callbacks redirect into the app (e.g. `/login`) with `?error=` and `errorCode=`, matching the mobile deep-link contract. Apps can read those params on a normal page instead of hitting raw JSON or a standalone HTML callback.
> 
> **`sendOAuthError`** now prefers that redirect for `platform: 'web'`, sets `Referrer-Policy: no-referrer`, and falls back to deprecated **`errorComponent`** then JSON if the URL cannot be built. A new **`platform: 'api'`** target keeps **initiation-time** failures (misconfigured provider, bad mobile `redirectUri`, missing `codeChallenge`) as JSON responses.
> 
> **`errorComponent`** is documented as deprecated; redirect wins when both are set. Invalid OAuth state user copy changes to *"Your sign-in session has expired…"* while **`invalid_state`** stays stable for branching.
> 
> OAuth shared code moves from **`oauth-common`** to **`oauthCommon`**, with **`buildOAuthErrorRedirect`** extracted and tests/docs updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98fb042dd20463cc6ddcca72b60d5509b38dc27e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/web/authentication/hooks.mdx
+++ b/docs/web/authentication/hooks.mdx
@@ -247,7 +247,7 @@ OAuth errors surface at the provider callback URL (`/api/_internal/auth/...`), w
 
 ### `oauthErrorRedirectUrl`
 
-Use `oauthErrorRedirectUrl` to send the browser back into your app when a web OAuth flow fails. It accepts a path or an absolute URL. The failure is appended as `?error=<message>&errorCode=<code>`, the same contract a mobile flow delivers on its deep link, so your app can show it on a real page.
+Use `oauthErrorRedirectUrl` to send the browser back into your app when a web OAuth flow fails. It accepts a path, resolved against your site URL (`MODELENCE_SITE_URL`), or an absolute URL. The failure is appended as `?error=<message>&errorCode=<code>`, the same contract a mobile flow delivers on its deep link, so your app can show it on a real page.
 
 ```typescript
 startApp({
--- a/docs/web/authentication/hooks.mdx
+++ b/docs/web/authentication/hooks.mdx
@@ -241,15 +241,41 @@ startApp({
 });
 ```
 
-## Error Rendering
+## Error Handling
 
-### `errorComponent`
+OAuth errors surface at the provider callback URL (`/api/_internal/auth/...`), which is a server route outside your client bundle. Nothing there can render your app's UI, so by default the user lands on a JSON body with no way back. Two options change that.
 
-Use `errorComponent` to customize how OAuth authentication errors are rendered. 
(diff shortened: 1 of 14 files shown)


eduard-piliposyan (changes requested):
- [ ] **Redirect crashes when `_system.site.url` is unset.** `sendOAuthError` casts `getConfig('_system.site.url')` to `string` and passes it as the base to `new URL(...)`. If the config is undefined and `oauthErrorRedirectUrl` is a relative path (e.g. `/login`), the constructor throws `Invalid URL` inside the error handler and the request ends as an unhandled 500 instead of a redirect. The comment says the site URL is always set by callback time, which holds for callback errors, but not for initiation-time errors: the 503 "not configured" responses in `google.ts` / `github.ts`, and the `redirectUri` / `codeChallenge` rejections in `resolveMobileRedirectRequest`, all run before `getRedirectUri`. Reachable in the `withoutRemoteServer` local variant without `MODELENCE_SITE_URL`. Suggest falling back to `getLocalSiteUrl()`, or catching the error and falling through to the JSON response with a logged warning.

- [ ] **Initiation-time errors for mobile clients now redirect to the web login page.** When `oauthErrorRedirectUrl` is set, a mobile app hitting the initiation endpoint with a disallowed `redirectUri` or a missing `codeChallenge` gets a 302 into the web app instead of a JSON 400. These errors have no outcome yet, so they take the web branch. Not a regression for existing users since it's opt-in, but either the docstring should state that the redirect also covers pre-state errors, or those paths should keep returning JSON.

Also, since this adds a public `AuthConfig` option and deprecates `errorComponent`, should we bump the `modelence` version in this PR?

omegascorp:
@eduard-piliposyan please check again

eduard-piliposyan (approved):

PR #362, opened 2026-09-02
fix: mobile oauth in iframe

omegascorp:
Fix mobile OAuth in iframe, this is reuired to make it property in the web preview

---

> [!NOTE]
> **Medium Risk**
> Touches authentication client flows and cross-window messaging; mistakes could break sign-in or weaken handoff validation, though changes are scoped to the optional popup path and include extensive tests.
> 
> **Overview**
> Adds **popup-based OAuth** for apps using the mobile `redirectUri` flow while running in a **cross-origin iframe** (embedded previews, sandboxes). When `configureClient`’s `openUrl` **returns the `window.open` reference** — e.g. `(url) => window.open(url)` — the opener keeps the PKCE verifier and listens for the exchange code over **`postMessage`**; the popup callback’s `loginWithOAuth` can **hand the code to the opener** and resolve to **`null`** so the session lands in the iframe, not partitioned popup storage.
> 
> New client APIs: **`resumeOAuthPopup`** (re-arm after HMR/preview reload; verifier still in `sessionStorage`) and **`cancelOAuthPopup`**, plus read-only **`hasPendingOAuthVerifier`**. **`signInWithOAuth`** now errors if the browser blocks the popup (`null` from `window.open`); shared redemption lives in **`redeemOAuthCode`**. Security checks in **`oauthPopupHandoff`**: peer identity via `event.source`, same-origin only, no session token on the wire.
> 
> **`ClientConfig.openUrl`** return type is **`unknown`** to allow returning a window. Docs cover iframe framing (`security.frameAncestors`), Google/GitHub sign-in, and React Native “Embedded previews and popups”. Package **0.24.1** with broad unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98587e726628d3b431e76b13eb9b5792cc0424cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


eduard-piliposyan (changes requested):
Here is a list of issues found after the review.

### High priority

**1. The session never reaches the embedding page, and there is no supported way to hand it over.**
`loginWithOAuth` runs in the popup, so `completeLogin` → `persistSession` writes the token to the *popup's* `localStorage` (or calls the app's `setAuthToken` in the popup's JS context) and `setCurrentUser` updates the *popup's* store. The iframe, whose storage is partitioned (the exact premise of this PR), stays signed out. The docs punt on this ("post the token to `window.opener`… then close the popup"), but that recipe is not implementable with the public API:
- `loginWithOAuth` returns the user, not the token.
- `modelence/client` exports no way to install a token and refresh the session store in the iframe: `setCurrentUser`, `initSession`, `reconcileSession`, `hydrateSession` are all internal.

As merged, the PR fixes the verifier half but the "web preview" flow still cannot complete end to end. Options, roughly in order of preference:
- Have the *opener* finish the login: the popup posts the `code` back over the same channel, the iframe (which still holds the verifier) calls `_system.user.loginWithOAuth` and `completeLogin` itself, then closes the popup. Solves this issue and removes the need for the verifier to leave the page at all.
- Or, drop the `postMessage` dependency entirely: let the opener redeem by verifier (server looks up the pending exchange entry whose challenge hash matches). This also removes the COOP fragility in #3.
- At minimum, export an `adoptSession({ authToken })`-style API and return the token from `loginWithOAuth`, so the documented recipe is actually possible.

**2. Any reload of the iframe during the round trip breaks the flow with a misleading error.**
The offer lives only in memory (`cancelActiveOffer` + the `opened` reference). If the embedded app reloads while the popup is at the provider (HMR / preview refresh, very common in the exact "web preview" scenario), nobody answers the request. The popup then waits the full 3 s and throws "This sign-in link is no longer valid", even though the verifier is still sitting in the iframe's `sessionStorage`. The opener-side design in #1 would fix this too (the iframe can re-arm on load when it finds a verifier in storage). Otherwise, document it.

**3. The feature depends on third parties not enforcing COOP, and fails silently when they do.**
`accounts.google.com` already sends `Cross-Origin-Opener-Policy: same-origin` in report-only mode on its sign-in pages. The moment Google (or GitHub) enforces it, `window.opener` is severed and every embedded sign-in degrades to a 3 s wait plus the generic "sign in again" error, with nothing in the console pointing at COOP. The docs `<Warning>` acknowledges this, but a core auth path should not hinge on a header we do not control. See the redeem-by-verifier option in #1; short of that, at least log a distinct console message when `requestVerifierFromOpener` returns `null` because `window.opener` is `null` on a page that arrived with `?code=`.

### Medium / docs

- **Framing prerequisite missing from the new docs section.** The Modelence server sets `X-Frame-Options: SAMEORIGIN` and `CSP frame-ancestors 'self'` by default (`packages/modelence/src/app/server.ts`). A reader following "Embedded previews and popups" alone will get a blocked frame before OAuth ever matters. Link to `security.frameAncestors`.
- **Section placement.** The only place this is documented is `docs/web/react-native.mdx`, but the feature is web-only (embedded previews). Plain web users running in an iframe will not find it; consider a short section in the web auth docs that points here.
- **Popup blocked → silent no-op.** `window.open` returns `null` when blocked; `signInWithOAuth` returns normally and nothing happens. Pre-existing, but now that returning the window is the documented pattern, consider throwing (or at least `console.error`) when `openUrl` returns `null` in a browser.

### Nits

- `packages/modelence/src/auth/client/index.ts`: the `loginWithOAuth` JSDoc has a stray hard wrap ("…cross-origin iframe. A plain web app\n * that calls…").
- `oauthPopupHandoff.ts` module comment says "the messages carry no nonce" as a design point; worth adding one sentence on why `event.source` identity is sufficient (WindowProxy identity is stable across navigations absent a BCG swap), since that is the whole security argument.

omegascorp:
@eduard-piliposyan feedback is applied, please check again

eduard-piliposyan (approved):
Looks good to me now.

PR #364, opened 2026-09-03
fix(cron): keep isRunning true after timeout to prevent re-run

Kathircpe:
The timeout scan previously cleared isRunning while the handler was still executing. This allowed the same job to be started again on the next tick, running two copies concurrently and clobbering state.

Now isRunning stays true until handleCronJobCompletion runs. A timedOutAt timestamp ensures captureError fires once, not every tick.

Followed from #363 .

---

> [!NOTE]
> **Medium Risk**
> Changes core cron scheduling and timeout behavior; incorrect logic could wedge jobs or duplicate runs, but the change aligns state with in-flight handlers.
> 
> **Overview**
> Fixes a bug where **cron jobs could run twice concurrently** after a timeout: the scheduler used to set `isRunning` to false when a run exceeded `timeout`, even though the handler was still executing, so the next tick could start another run.
> 
> **`timedOutAt`** is added to job state so `captureError` for the timeout fires **once** per stuck run instead of on every tick. `isRunning` stays true until **`handleCronJobCompletion`** runs when the handler finishes or errors.
> 
> Tests cover single timeout reporting and that the handler is not invoked again while a timed-out run is still in flight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8e56d113919d5b7453330bb7db851995a3bef86. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/cron/jobs.test.ts
+++ b/packages/modelence/src/cron/jobs.test.ts
@@ -250,7 +250,7 @@ describe('cron/jobs', () => {
     expect(transaction.end).toHaveBeenCalledWith('error');
   });
 
-  test('cron loop handles and captures timeout when job runs past its timeout limit', async () => {
+  test('cron loop captures timeout when job runs past its timeout limit', async () => {
     mockCaptureError.mockClear();
     let resolveHandler: () => void;
     const handlerPromise = new Promise<void>((resolve) => {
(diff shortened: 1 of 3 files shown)


Kathircpe:
@omegascorp take a look at this.

packages/modelence/src/cron/jobs.test.ts
```diff
@@ -285,4 +285,71 @@ describe('cron/jobs', () => {
...
+    vi.spyOn(Date, 'now').mockReturnValue(startTs);
+    await intervalCallback?.();
+
+    (Date.now as Mock).mockReturnValue(startTs + 6000);
+    await intervalCallback?.();
+
+    (Date.now as Mock).mockReturnValue(startTs + 12000);
+    await intervalCallback?.();
+
+    expect(mockCaptureError).toHaveBeenCalledTimes(1);
```
omegascorp, line 318: This one passes on the unfixed code, for a subtler reason than the other test.

The tick at `+6000` does exercise the `timedOutAt` guard as intended. But the tick at `+12000` is past the 10s interval, so on old code the job has already been re-admitted — `runCronJob` resets `startTs` to the current mocked time, and `startTs + timeout` is no longer `< now`. So the third tick doesn't fire a second `captureError` on old code either, and the assertion lands on 1 by accident rather than because the guard held.

To make it test the guard, keep both post-timeout ticks inside the window where the job is timed out but not yet due to restart — e.g. `+6000` and `+9000`, both past `timeout: 5000` but under `interval: 10000`. Then old code fires twice and the fix fires once.

Asserting `expect(handler).toHaveBeenCalledTimes(1)` alongside the `captureError` count would also pin down that no restart happened, which is what makes the single capture meaningful.

packages/modelence/src/cron/jobs.test.ts
```diff
@@ -285,4 +285,71 @@ describe('cron/jobs', () => {
...
+
+    const startTs = Date.now();
+    vi.spyOn(Date, 'now').mockReturnValue(startTs);
+    await intervalCallback?.();
+
+    (Date.now as Mock).mockReturnValue(startTs + 6000);
+    await intervalCallback?.();
+
+    expect(handler).toHaveBeenCalledTimes(1);
+
```
omegascorp, line 351: This is the test that should have caught the bug, and it doesn't — it passes unchanged against pre-fix `b5106e18`.

At `startTs + 6000` the job is past its 5s timeout, so old code sets `isRunning = false`. But `scheduledRunTs` is still `startTs + 10000`, so `state.scheduledRunTs <= now` is false and nothing restarts on this tick regardless. The assertion can't distinguish the two implementations.

Two changes needed:

1. **Add a third tick past the interval**, which is where the duplicate actually appears:
   ```ts
   (Date.now as Mock).mockReturnValue(startTs + 11000);
   await intervalCallback?.();
   ```
2. **Drain microtasks before asserting.** `tickCronJobs` calls `runCronJob` without awaiting it, so the handler call isn't observable on the same turn:
   ```ts
   await new Promise((resolve) => setImmediate(resolve));
   ```

With both, old code reaches `handler` twice and this branch keeps it at 1 — I confirmed that locally on the exact commits. Same microtask drain is worth adding after the `+6000` tick so the intermediate state is settled when you assert on it.

omegascorp (changes requested):
The production change is right and I want it — this is the prerequisite for #363, and the bug is real. I reproduced it against pre-fix `main`: with `timeout: 5000` and `interval: 10000`, ticks at t+0 / t+6s / t+11s start a **second copy** of the handler, and one more every interval after that. Your three-line fix in `jobs.ts` brings that to exactly 1. Clearing `timedOutAt` in both `runCronJob` and `handleCronJobCompletion` is the right pair — a job that eventually finishes can still time out on a later run, no leak.

**But both new tests pass against the unfixed code, so neither would have caught the bug.**

I checked this directly: reverted `jobs.ts` and `types.ts` to `b5106e18` (pre-fix), left your test file untouched, ran it — 11/11 green. Full suite is 1078 passing and `tsc --noEmit` is clean on your branch, so this is only about the tests proving what they claim.

The cause is the same in both: they stop at `startTs + 6000`, which is past the 5s timeout but **before** the 10s interval. `scheduledRunTs` is still in the future, so the admission check `state.scheduledRunTs <= now` fails on its own — the old code's `state.isRunning = false` is invisible at that instant. The duplicate run only surfaces on the next tick past the interval.

Details inline on each test. Worth re-running both against `b5106e18` before pushing — if they still pass, they aren't testing the fix yet.

One more thing: this branch is based on `b5106e18`, which is 3 commits behind `main` (now `1673fdda`). Nothing conflicts in `src/cron/`, but please rebase before merge.

omegascorp:
Hi @Kathircpe I'm leaving this for another week. I'm checking community PRs on Mondays

PR #365, opened 2026-09-03
fix: capture method errors via captureError before ending transaction

Kathircpe:
## Summary
runMethod / runLiveMethod ended the transaction as 'error' but never reported the error itself, so failures were invisible in APM. This calls captureError() (normalizing non-Error throws) before transaction.end('error'), mirroring the cron-job pattern, and extends the methods/index tests to assert the capture on both paths.

---

> [!NOTE]
> **Low Risk**
> Observability-only change; errors are still rethrown and transaction behavior is unchanged aside from associating captured errors.
> 
> **Overview**
> Method handlers already ended telemetry transactions as **error** on failure but did not record the exception, so APM showed failed transactions without the underlying error. **`runMethod`** and **`runLiveMethod`** now call **`captureError`** in their catch blocks (coercing non-`Error` throws to `Error`) before **`transaction.end('error')`**, matching the existing cron-job pattern.
> 
> Tests mock **`captureError`** and assert it runs alongside the error transaction end for both standard and live method execution paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 178f2c3bfa0e523a82bdff11c12a18fdf9e0b8cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/methods/index.test.ts
+++ b/packages/modelence/src/methods/index.test.ts
@@ -4,6 +4,7 @@ import type { Handler } from './types';
 const transactionEnd = vi.fn();
 const mockRequireServer = vi.fn();
 const mockStartTransaction = vi.fn();
+const mockCaptureError = vi.fn();
 const mockRequireAccess = vi.fn();
 
 vi.doMock('../utils', () => ({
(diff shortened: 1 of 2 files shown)


omegascorp (approved):
Looks good, not merging yet, will do once 0.24.0 is released

PR #366, opened 2026-09-03
MOD-618: Claude Code plugin: OAuth-only auth, CLI install from setup

artahian:
## Summary
- Plugin 0.3.0: the `modelence` MCP server authenticates with OAuth only. The service-token `headersHelper` and `scripts/mcp-auth.js` are removed (the Agent Plugins standard forbids header expansion), and the server URL is the literal `https://cloud.modelence.com/mcp` — the `${MODELENCE_MCP_URL:-…}` override broke Claude Desktop's Connect button.
- `modelence setup` installs the plugin: after writing `.claude/settings.json` it asks `Install the Modelence plugin for Claude Code? (Y/n)` and runs `claude plugin marketplace add` + `claude plugin install … --scope project`. Skipped with `--token` or without a TTY; a missing CLI or a failed install prints the two commands and never fails setup.
- Docs (`ai-coding-agents`) drop all per-client GUI instructions in favour of CLI-only install/reinstall steps and the `/mcp` sign-in, so a user whose plugin got removed can have Claude Code look it up. README links to the docs instead of duplicating them.
- Every `npx modelence setup` reference becomes `npx modelence@latest setup`.

Studio keeps accepting the old service-token header so existing 0.2.0 installs continue to work (third-party marketplaces don't auto-update).

## Test plan
- [ ] `npx modelence@0.24.0-mcp-oauth.dev.0 setup` in a scratch project: prompt shown, plugin installed, project `.claude/settings.json` correct
- [ ] Same with `--token`: no prompt, no install
- [ ] Claude Desktop: Connect on the `modelence` server completes the OAuth flow
- [ ] Docs build, `#claude-code-plugin` anchor resolves

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes how developers authenticate to the cloud MCP (OAuth required for 0.3.0) and how setup mutates local Claude Code config; backward compatibility for older plugin installs is noted as a server-side concern outside this diff.
> 
> **Overview**
> **Claude Code plugin 0.3.0** drops automatic service-token MCP auth: `scripts/mcp-auth.js` and the `headersHelper` in `.mcp.json` are removed, and the cloud MCP URL is fixed to `https://cloud.modelence.com/mcp` (no `${MODELENCE_MCP_URL}` override). Users authenticate the `modelence` MCP server once via OAuth in Claude Code (`/mcp` → Authenticate). Plugin guidance now treats MCP as **org-wide** (tools take `environmentId`; agents read `MODELENCE_ENVIRONMENT_ID` from `.modelence.env` for the connected project).
> 
> **`npx modelence@latest setup`** (modelence **0.24.2**) merges marketplace + enabled-plugin entries into `.claude/settings.json`, then optionally runs `claude plugin marketplace add` and `claude plugin install modelence@modelence --scope project` when the CLI is on PATH and the run is interactive. CI/`--token` and non-TTY runs skip the prompt; failures print manual install steps without failing setup.
> 
> Docs and in-app copy are aligned: install/reinstall and sign-in live on **ai-coding-agents**, README points there, and setup command strings use **`npx modelence@latest setup`** everywhere. After reconnecting an environment, messaging tells users to restart the dev server and start a **new agent session** instead of reconnecting MCP.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5662c051da7b3e7c8384e08388b5cf64d8f2e8b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


omegascorp:
@artahian I haven't released 0.24.0 yet; it contains an issue with mobile OAuth, which is fixed here: https://github.com/modelence/modelence/pull/362 

I'm waiting on approval to release it

packages/modelence/src/bin/setup.ts
```diff
@@ -116,60 +121,91 @@ const CLAUDE_PLUGIN_ID = 'modelence@modelence';
...
+*/
+async function offerClaudePluginInstall(): Promise<void> {
+  const probe = runClaude(['--version']);
+  if (probe.error || probe.status !== 0) {
+    console.log(CLAUDE_PLUGIN_INSTALL_HINT);
+    return;
+  }
+
+  const answer = await ask('Install the Modelence plugin for Claude Code? (Y/n) ');
+  if (answer && answer.toLowerCase() !== 'y') {
```
omegascorp: Could we also accept `yes`? Because `yes` currently means no, people might type it automatically

omegascorp:
A few more things beyond the prompt-parsing one:

### Existing projects never get `extraKnownMarketplaces` back

`packages/modelence/src/bin/setup.ts`:

```ts
// An explicit `false` is the user's opt-out — leave it alone.
if (settings.enabledPlugins?.[CLAUDE_PLUGIN_ID] !== undefined) {
  return;
}
```

The comment says "an explicit `false`", but the check is `!== undefined`, so it also returns early when the value is `true`. A settings.json with `enabledPlugins: { "modelence@modelence": true }` and no `extraKnownMarketplaces` therefore never gets the marketplace declaration added.

That's exactly the shape a hand-edited file from the old docs has, and it's also what `claude plugin install --scope project` leaves behind — I checked, and it writes `enabledPlugins` to the project file while declaring the marketplace in *user* settings. So the combination is common, not hypothetical.

Per the comment a few lines above, without the marketplace alongside `enabledPlugins` Claude Code can't resolve the plugin. Teammates cloning such a repo get an unresolvable plugin entry, and re-running setup won't repair it.

The two keys are independent, so they should be gated independently:

```ts
settings.extraKnownMarketplaces = { ...settings.extraKnownMarketplaces, ...CLAUDE_MARKETPLACES };
if (settings.enabledPlugins?.[CLAUDE_PLUGIN_ID] === undefined) {
  settings.enabledPlugins = { ...settings.enabledPlugins, [CLAUDE_PLUGIN_ID]: true };
}
```

Only `enabledPlugins` carries the opt-out — declaring where a marketplace lives isn't something a `false` opts out of.

### `marketplace add` result is discarded

```ts
runClaude(['plugin', 'marketplace', 'add', CLAUDE_MARKETPLACE_REPO]);
const install = runClaude(['plugin', 'install', CLAUDE_PLUGIN_ID, '--scope', 'project']);
```

If the `marketplace add` fails for a reason other than "already added" — network, auth — the install then fails with a message that doesn't point at the actual cause. Cheap to check its status and fold a failure into `CLAUDE_PLUGIN_INSTALL_HINT`.

### The bundled-CLI note reads as reassurance

`docs/web/ai-coding-agents.mdx`:

> The desktop app and the IDE extensions bundle their own private copy, so if `claude` isn't found, install the CLI from code.claude.com first.

The actionable point is the opposite of how this skims: the bundled copy isn't on PATH, so `npx modelence setup` can't use it. Someone who reads "I have the desktop app, so I have a copy" will skip the install and then wonder why setup printed the manual steps. Suggest something like:

> The desktop app and IDE extensions ship a private copy that isn't on your PATH, so `npx modelence setup` can't use it — install the standalone CLI from code.claude.com if `claude` isn't found.

---

Two smaller ones, take or leave:

- Ordering churn: `ensureClaudePluginEnabled()` writes `.claude/settings.json`, then `claude plugin install --scope project` rewrites it. Content is preserved, but the keys come back reordered, so every setup run produces a spurious diff on a file the docs now tell people to commit.
- In `ensureClaudePluginEnabled`, a read failure and a parse failure share one `catch`, so an `EACCES` on the file would print "Could not parse …" and send someone hunting for a syntax error that isn't there.

omegascorp (approved):
Looks good

PR #367, opened 2026-09-04
fix(rate-limit): use atomic findOneAndUpdate to prevent TOCTOU race condition (#353)

### LINKED ISSUE #353: Security: TOCTOU race condition in consumeRateLimit allows rate-limit bypass during concurrent bursts

ankit-0903:
Fixes #353

### Summary of Changes
- Refactored `checkRateLimitRule` in `packages/modelence/src/rate-limit/rules.ts` to perform atomic MongoDB updates using `findOneAndUpdate` with `$inc: { windowCount: 1 }`.
- Eliminates pre-increment `findOne` read operations for active windows, preventing concurrent requests from bypassing rate limits via TOCTOU race conditions.
- Uses conditional matching on `windowStart` during window transitions to safely handle concurrent window shifts.
- Updated unit test suite in `rules.test.ts` for atomic operations and window transitions. All unit tests, formatting, and linting pass cleanly.

---

> [!NOTE]
> **Medium Risk**
> Changes core rate-limit enforcement and Mongo update sequencing under concurrency; bugs could mis-count or block legitimate traffic, though fail-closed and rollback aim for stricter—not looser—limits.
> 
> **Overview**
> Replaces the **read-then-`upsertOne`** flow in `checkRateLimitRule` with a **three-step retry loop** that increments and shifts windows via **`findOneAndUpdate`**, **`findOneAndUpsert`**, and conditional **`windowStart`** matching so concurrent requests cannot slip past limits between check and write.
> 
> After each optimistic **`$inc`**, the limit is evaluated on the returned document; rejections **roll back** with **`updateOne`** **`$inc: -1`**. Window rollovers use atomic **`$set`** only when the stored **`windowStart`** still matches, and the loop retries when another request already advanced the window. If three attempts still cannot settle, it **logs and fails closed** (throws rate limit exceeded).
> 
> Tests now mock the new DB methods and add coverage for window shifts, concurrent **`windowStart`** races, and exhausted retries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65061e5d38c0105dce5592cd99c84fa7ea21eecd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Kathircpe (commented):
Verdict: Correct direction but still fail-open.

Original bug is real (rules.ts:53-88: findOne → local count >= limit → upsertOne). PR replaces it with atomic findOneAndUpdate / findOneAndUpsert + retry. Hot path fixed, but retry-exhaustion path reintroduces bypass.

packages/modelence/src/rate-limit/rules.ts
```diff
@@ -50,88 +50,116 @@ async function checkRateLimitRule(rule: RateLimitRule, value: string, createErro
...
+    const updatedDoc = await dbRateLimits.findOneAndUpdate(
+      { ...filter, windowStart: currentWindowDate },
+      { $inc: { windowCount: 1 } },
+      { returnDocument: 'after' }
+    );
+
+    if (updatedDoc) {
+      const prevWeight = 1 - (now - currentWindowStart) / rule.window;
+      const count = Math.round(updatedDoc.windowCount + updatedDoc.prevWindowCount * prevWeight);
+      if (count > rule.limit) {
```
omegascorp, line 69: This block (compute weighted count → if over limit, decrement and throw → otherwise return) is repeated verbatim three times in the loop. Suggest pulling it into a helper so each step reads as `return acceptOrReject(doc)`:

```ts
const acceptOrReject = async (doc: (typeof dbRateLimits)['Doc']) => {
  const prevWeight = 1 - (now - currentWindowStart) / rule.window;
  const count = Math.round(doc.windowCount + doc.prevWindowCount * prevWeight);
  if (count > rule.limit) {
    // Compensate the optimistic increment. If a window shift raced us, this
    // matches nothing and the count leaks into prevWindowCount — stricter, not looser.
    await dbRateLimits.updateOne(
      { ...filter, windowStart: currentWindowDate },
      { $inc: { windowCount: -1 } }
    );
    throw createRateLimitError();
  }
};
```

Defined inside the loop body it closes over `now`, `currentWindowStart`, and `currentWindowDate` for that attempt, so the three call sites become `await acceptOrReject(updatedDoc); return;` etc.

omegascorp (changes requested):
Thanks @ankit-0903 — this is the right fix for #353, and the head revision (da0225c) holds up well. I checked the algorithm against the earlier Bugbot findings and @Kathircpe's review, and they're all addressed now: the `>=` guard in step 2 prevents a stale request from rolling a newer window back, the decrement on rejection keeps `limit: 1` from wedging, and exhausting the retries fails closed. I also confirmed the budget is unchanged from the old code — the old pre-increment `>=` and the new post-increment `>` (with decrement on reject) are equivalent, including through the `Math.round` on the weighted previous-window term. Tests pass locally and the module typechecks.

Two things before we merge:

### 1. Rebase onto `main`

#355 landed and the branch now conflicts on `rules.ts`. Your branch already contains the identical `filter` const, so the resolution should be mechanical — keep your version of the function.

### 2. Extract the repeated rejection block

The `compute count → decrement if over → throw` sequence is copied three times (step 1, the shift path, and the upsert path). See the inline comment for a suggested helper. This is the one structural change I'd like before merging — the loop is doing subtle CAS work, and having the accept/reject logic in one place makes it much easier to reason about.

### Non-blocking, for the record

- A rejected request now costs two writes (increment + compensating decrement) where the old code cost one read. Under a sustained attack that roughly doubles write load on `_modelenceRateLimits`. Fine for now; I'll open a follow-up issue so we don't lose track of it.
- If a window shift races between the increment and the decrement, the decrement matches nothing and one count leaks into the next window's `prevWindowCount`. That errs toward stricter limiting, so it's safe — just worth a one-line comment near the decrement so the next reader doesn't think it's a bug.

Once those are in I'll merge and close #353.

ankit-0903:
Thanks for the thorough review @omegascorp!
I've addressed both items:

   - *Rebased onto main* - resolved the conflict with #355 by keeping our
   version of the function.
   - *Extracted acceptOrReject helper* - the repeated accept/reject block
   across all three steps is now centralized in a single helper defined inside
   the loop, with a comment near the compensating decrement explaining that a
   lost race leaks into prevWindowCount (stricter, not looser).

All 9 tests pass. Ready for another look!

packages/modelence/src/rate-limit/rules.ts
```diff
@@ -57,80 +57,99 @@ async function checkRateLimitRule(rule: RateLimitRule, value: string, createErro
...
+  for (let attempt = 0; attempt < 3; attempt++) {
+    const now = Date.now();
+    const currentWindowStart = Math.floor(now / rule.window) * rule.window;
+    const currentWindowDate = new Date(currentWindowStart);
+    const expiresAt = new Date(currentWindowStart + rule.window + rule.window);
+
+    const acceptOrReject = async (doc: { windowCount: number; prevWindowCount: number }) => {
+      const prevWeight = 1 - (now - currentWindowStart) / rule.window;
+      const count = Math.round(doc.windowCount + doc.prevWindowCount * prevWeight);
+      if (count > rule.limit) {
```
omegascorp, line 71: Worth a short comment noting this comparison is intentionally post-increment: the doc has already been `$inc`'d by one at this point, so `>` here is equivalent to the old pre-increment `>= rule.limit`. I confirmed the admitted counts are identical to `main` for limits 1/2/3/5, but as written it reads like an off-by-one and someone will eventually "fix" it back to `>=`.

packages/modelence/src/rate-limit/rules.ts
```diff
@@ -57,80 +57,99 @@ async function checkRateLimitRule(rule: RateLimitRule, value: string, createErro
...
+        // If not newly inserted, a concurrent request created it; retry loop to increment via step 1
+        continue;
+      }
+      await acceptOrReject(upsertedDoc);
+      return;
+    }
+  }
+
+  // Fail-closed if retry attempts are exhausted under extreme concurrency.
+  throw createRateLimitError();
```
omegascorp, line 159: Could we log before this throw? This is the one path that rejects a legitimate user rather than an abusive one, and right now it's indistinguishable from a normal rate-limit rejection at the call site. Exhausting all 3 attempts should be very rare, so if it does start firing in production we'd want to be able to see it.

omegascorp (commented):
Verified the fix does what it says: under a concurrent burst `main` admits the whole burst (20/20 against a limit of 5), this branch admits exactly 5. Sequential behaviour is unchanged across limits 1/2/3/5, sustained multi-window traffic, and counter drift under heavy rejection.

This matters more than the issue suggests, because the auth defaults in `src/auth/user.ts` ship enabled: `verification` is `limit: 1` per 60s (the entire limit fits in the TOCTOU gap), and `passwordReset`/`magicLink` are `limit: 5` per email — so failing open there means sending N emails to a third party.

Two small asks below, both comment/logging only — no logic changes. The three Bugbot comments above are stale (they target df9f077/f2cd29c/a32c09b; head is e8d6254) and are addressed by the `>= currentWindowStart` guard and the rollback.

PR #368, opened 2026-09-04
feat: route server error logs to Elastic APM when telemetry is enabled

Kathircpe:
## Summary
Replace unconditional console.error in handleMethodError (app/server.ts) and the route handler (routes/handler.ts) with a centralized telemetry.reportError sink: APM capture with method/route context when telemetry is on, console otherwise (with APM-failure fallback so responses never break). Non-Error throws keep a searchable message plus the raw value on cause.

## Note 
- Object.assign(new Error(msg), { cause }) vs new Error(msg, { cause: error }) — latter is idiomatic ES2022, non-enumerable cause. Current works, just unidiomatic.
- cron/jobs.ts:114,149 still uses raw captureError — fine (it already branches internally), but you now have two public APIs (captureError + reportError). Consider migrating cron to reportError later for one sink.

---

> [!NOTE]
> **Low Risk**
> Observability-only change: HTTP status bodies and client-visible errors are unchanged; risk is limited to where errors are logged when telemetry is enabled.
> 
> **Overview**
> Introduces **`telemetry.reportError`** as the shared server-side error sink for method and route handlers. When Elastic APM telemetry is on, failures are sent to APM once with a **context string** (e.g. method name or route path); when it is off, behavior stays **`console.error`** with that context. The helper **never throws** and falls back to the console if APM fails, so Express responses are unchanged.
> 
> **`handleMethodError`** in `app/server.ts` and the generic **`catch`** in `routes/handler.ts` now call `reportError` instead of raw `console.error` (including 5xx `ModelenceError`, Zod parse failures, and unknown throws). Non-`Error` values are wrapped for APM with a searchable message and **`cause`** preserved.
> 
> Tests cover the new telemetry behavior and wire **`reportError`** mocks in server and route handler tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0d36529a9a0f145c9ca08402b094b7f72a657c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/app/server.test.ts
+++ b/packages/modelence/src/app/server.test.ts
@@ -56,6 +56,7 @@ const mockCreateRouteHandler =
 const mockGoogleAuthRouter = vi.fn();
 const mockGithubAuthRouter = vi.fn();
 const mockLogInfo = vi.fn();
+const mockReportError = vi.fn();
 const mockGetSecurityConfig = vi.fn();
 const mockGetWebsocketConfig = vi.fn();
 const mockExpressJson = vi.fn();
@@ -106,6 +107,7 @@ vi.doMock('@/auth/providers/github', () => ({
 
 vi.doMock('@/telemetry', () => ({
   logInfo: mockLogInfo,
+  reportError: mockReportError,
 }));
 
 vi.doMock('./securityConfig', () => ({
--- a/packages/modelence/src/app/server.ts
+++ b/packages/modelence/src/app/server.ts
@@ -4,7 +4,7 @@ import { runMethod } from '@/methods';
 import { getResponseTypeMap, sanitizeResult } from '@/methods/serialize';
 import { createRouteHandler } from '@/routes/handler';
 import { HttpMethod } from '@/server';
-import { logInfo } from '@/telemetry';
+import { logInfo, reportError } from '@/telemetry';
 import cookieParser from 'cookie-parser';
 import express, { Request, Response } from 'express';
 import http from 'http';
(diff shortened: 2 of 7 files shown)


Kathircpe:
@omegascorp take a look at this.

omegascorp (commented):
Thanks for the PR. I don't think we need this change anymore, so I'd suggest closing it.

**Why:** #365 (merged this morning) already makes `runMethod` call `captureError` on every thrown error, and `captureError` does exactly what `reportError` does here: APM when telemetry is on, `console.error` otherwise. `handleMethodError` then runs on the same error right after `runMethod` rethrows it.

So on top of current `main`, this PR would give us:

| Path | Telemetry off | Telemetry on |
|---|---|---|
| Method error (main today) | console twice | APM once + console once |
| Method error (with this PR) | console twice | APM **twice** (one without context, one with) |
| Route error (with this PR) | console once | APM once |

The method path doesn't get better and gains a duplicate APM capture. Only the route handler benefits, and it can get the same effect with the existing `captureError`.

A few other notes:
- `reportError` duplicates `captureError`; as your own note says, that leaves two public APIs for the same sink. If we want a context string, an optional second argument on `captureError` is enough.
- The tests mock the telemetry module, so they can't catch the double capture.
- The `captureError` in `runMethod` (from #365) currently captures 4xx `ModelenceError` and Zod validation failures too. That's worth a separate follow-up so we only report actual server errors.

If you'd like to pursue this, a smaller PR that picks a single capture point per request (probably `handleMethodError`, since it already knows 5xx vs client errors) and drops the extra `captureError` in `runMethod` would be welcome.

omegascorp:
@Kathircpe I'm closing the PR as it's not a required change

PR #369, opened 2026-09-05
fix(auth): replace console.error with captureError in resetPassword landing handler (#334)

### LINKED ISSUE #334: Telemetry: Replace console.error with captureError in resetPassword landing handler
### Description
In `packages/modelence/src/auth/resetPassword.ts` around line 192, `console.error` is used to log errors in the password reset landing handler. 

Most other production error paths were recently updated to use structured logging with `captureError` from `@/telemetry`, but this landing handler seems to have been missed.

### Expected Behavior
Errors in `handleResetPasswordLanding` should be captured using `captureError` so they show up properly in APM telemetry sinks.

### Proposed Solution
Import `captureError` from `@/telemetry` and replace the `console.error` call in `handleResetPasswordLanding`.

ankit-0903:
Fixes #334

### Summary
Replaces raw `console.error` with `captureError` from `@/telemetry` in `handleResetPasswordLanding` (`packages/modelence/src/auth/resetPassword.ts`), bringing error reporting in the password reset landing route into alignment with APM structured telemetry sinks used across other production error paths.

### Changes
- **`packages/modelence/src/auth/resetPassword.ts`**:
  - Imported `captureError` from `@/telemetry`.
  - Replaced `console.error` with `captureError(err)` in the `catch` block of `handleResetPasswordLanding`.
- **`packages/modelence/src/auth/resetPassword.test.ts`**:
  - Mocked `@/telemetry`'s `captureError`.
  - Added assertions in `handleResetPasswordLanding` tests to verify that `captureError` is called when invalid, expired, or missing reset tokens are handled.

### Verification
- **Vitest**: All 36 tests in `src/auth/resetPassword.test.ts` passed.
- **ESLint & Prettier**: Passed with 0 errors.

---

> [!NOTE]
> **Low Risk**
> Observability-only change on an auth error path; user-facing redirects and messages are unchanged.
> 
> **Overview**
> Password reset **landing route** failures (invalid, expired, or missing token) still redirect users with the same friendly message, but the server now sends the underlying error to **structured telemetry** via `captureError` instead of `console.error`.
> 
> The catch block in `handleResetPasswordLanding` normalizes thrown values to an `Error` before reporting. Tests mock `@/telemetry` and assert `captureError` runs on the three error redirect paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 309fc2d4774fdc8b5913af4ecfe5650b65219fc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/modelence/src/auth/resetPassword.test.ts
+++ b/packages/modelence/src/auth/resetPassword.test.ts
@@ -33,6 +33,11 @@ const mockConsumeRateLimit = vi.fn();
 const mockGetConfig = vi.fn();
 const mockInvalidateAllUserSessions = vi.fn();
 const mockMagicLinkTokensDeleteMany = vi.fn();
+const mockCaptureError = vi.fn();
+
+vi.doMock('@/telemetry', () => ({
+  captureError: mockCaptureError,
+}));
 
 vi.doMock('./session', () => ({
   invalidateAllUserSessions: mockInvalidateAllUserSessions,
@@ -1236,6 +1241,11 @@ describe('auth/resetPassword', () => {
       const result = await handleResetPasswordLanding(makeParams('bad-token', cookie) as never);
 
       expect(cookie).not.toHaveBeenCalled();
+      expect(mockCaptureError).toHaveBeenCalledWith(
+        expect.objectContaining({
+          message: 'This password reset link is invalid or has expired.',
+        })
+      );
       expect(result?.status).toBe(302);
       expect(result?.redirect).toContain('status=error');
       expect(result?.redirect).toContain(friendlyParam);
@@ -1253,6 +1263,11 @@ describe('auth/resetPassword', () => {
       const result = await handleResetPasswordLanding(makeParams('expired', cookie) as never);
 
       expect(cookie).not.toHaveBeenCalled();
+      expect(mockCaptureError).toHaveBeenCalledWith(
+        expect.objectContaining({
+          message: 'This password reset link is invalid or has expired.',
+        })
+      );
       expect(result?.redirect).toContain('status=error');
       expect(result?.redirect).toContain(friendlyParam);
     });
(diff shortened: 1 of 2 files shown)


omegascorp:
Thanks for putting this together, and for adding test coverage along with it — the underlying goal here is right, and auth error paths genuinely should be reaching APM rather than stdout.

That said, I'm going to close this one, for a few reasons:

**The premise in #334 isn't accurate.** The issue says most other production error paths were recently updated to use `captureError`. In practice `captureError` has exactly one production call site in the package — `cron/jobs.ts`, added in #337 — while there are 66 `console.error` calls in non-test source, 13 of them in `auth/` alone.

**It breaks an existing symmetry.** `handleResetPasswordLanding` and `handleMagicLinkLanding` are structurally identical catch blocks, down to the same comment. Converting one and not the other introduces an inconsistency where there currently isn't one.

**It drops log context.** `console.error('Error handling password reset landing:', error)` becomes `captureError(err)`, so the message identifying which handler failed is gone. Since `captureError` falls back to a bare `console.error(error)` when telemetry is disabled, that's a strict loss for self-hosted users.

**One test assertion is a bit misleading.** The invalid/expired tests assert `captureError` was called with `message: 'This password reset link is invalid or has expired.'`. That passes only because the internal `throw` happens to use the same string as the user-facing message — it reads as though it's checking the redirect message, and it pins an internal string that we'd want to change if we ever made that error more diagnostic.

I'm keeping #334 open and rescoping it to cover the whole `auth/` directory — `magicLink.ts`, `resetPassword.ts`, `verification.ts`, and the OAuth providers — as one deliberate pass, which likely also wants a message-carrying variant of `captureError` so context isn't lost. If you're up for taking that on, it would be very welcome.

PR #384, opened 2026-09-14
fix: close TOCTOU race in consumeRateLimit

### LINKED ISSUE #353: Security: TOCTOU race condition in consumeRateLimit allows rate-limit bypass during concurrent bursts

Abel-Guta:
## Summary

Closes #353. `checkRateLimitRule` read the current window count via `findOne`, decided in JS whether the limit was exceeded, and only then wrote the increment via `upsertOne`. Under concurrent burst traffic, multiple requests could all read a stale sub-limit count before any of them had written its increment, so all of them would pass — the rate limit bypasses entirely.

## What changed

Replaced the read-then-write with a single atomic `findOneAndUpdate` using an aggregation-pipeline update, so the window-rotation logic (same window / previous window / stale) runs server-side against whatever the document's actual state is at write time, instead of a value read a moment earlier in JS. The post-increment count is then checked against the limit.

## Trade-off worth flagging

A request that arrives after the limit is already exceeded now still increments `windowCount` (previously it only read and never wrote once over the limit), so `windowCount` can grow past the limit within a window. This doesn't change allow/deny behavior — it's the standard "increment then check" pattern most atomic rate limiters use — but it does mean the stored count for an over-limit window is no longer a clean "requests actually let through" number. Happy to add a cap if that's preferred.

## Validation

- `npm run test` (`vitest run`) — 9/9 passing, including 6 rewritten tests for `rate-limit/rules.ts` that mock `findOneAndUpdate` directly and assert on post-increment count
- `npx tsc --noEmit` — clean
- `npx eslint` on both changed files — clean

## Scope

Touches only `packages/modelence/src/rate-limit/rules.ts` and its test file.

---

> [!NOTE]
> **Medium Risk**
> Changes core rate-limit enforcement under concurrency; behavior is stricter and more correct, but rejected requests still increment stored counts and brief bursts may slightly exceed the limit.
> 
> **Overview**
> **Fixes a concurrency race in `consumeRateLimit`** where `findOne` plus a separate `upsertOne` let burst traffic bypass limits when multiple requests read the same stale count before any write.
> 
> `checkRateLimitRule` now uses one **`findOneAndUpdate`** with an **aggregation-pipeline update** (`upsert: true`, `returnDocument: 'after'`) so window rotation and increment run atomically in MongoDB. Allow/deny uses the **post-increment** weighted count (`count > limit` instead of pre-check `>=`), including the request that lands exactly on the limit.
> 
> The old JS `getCount` helper is removed. Tests mock `findOneAndUpdate` and cover atomicity, boundary allow/deny, custom errors, and `prevWindowCount` weighting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88c045472fd4f387e0e056ecd419cc634b682d32. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Abel-Guta:
Closing this - just found #367 already has this fixed and is nearly through review (thanks for the thorough back-and-forth there, @omegascorp / @ankit-0903). Didn't mean to land-grab; should have checked for an existing PR before opening this one. Sorry for the noise.

PR #385, opened 2026-09-15
feat: generic node deploy

omegascorp:
## Summary

`modelence deploy` now works for any Node.js app, from any directory, with no flags.

Before, deploy meant "build a Modelence app locally and upload `.modelence/build`", and it required `-a` and `-e`. Now the default is a remote build from source, the target is picked in the browser the first time, and the login is remembered.

## New commands

- **`modelence init`** — detects how the project builds and runs, writes it to `modelence.json` with a `$schema` link so editors autocomplete it, and prints the plan plus any notes. Detection still runs on every deploy for whatever the file leaves out.
- **`modelence logout`** — forgets the saved login (all hosts, or one with `-h`).

## `modelence deploy`

- `-a/-e` are optional. Target resolution: flags → the app recorded in `.modelence/project.json` → the deploy target recorded there → the browser picker. Once picked, it is written back to `.modelence/project.json` (committed; no credentials in it).
- Auth: `MODELENCE_TOKEN` → a cached 30-day token in `~/.modelence/auth.json` (mode `0600`, per host) → the browser. If Studio rejects the cached token mid-deploy, the CLI re-authorizes once and **resumes following the same build** rather than deploying twice.
- Source is packed with `git ls-files` inside a repo, so the upload matches what a push would carry; outside git it walks with a fixed exclusion list. `.env` files, `.modelence.env`, and `.npmrc`/`.yarnrc` containing literal auth tokens are dropped either way, and the CLI prints how many files it excluded. 200 MB cap.
- Progress is followed to the end: phase transitions, streamed build logs, the site URL, and — when the rollout fails — the last output of the containers that failed to start. Non-zero exit on failure. It also waits out a freshly created environment that is still provisioning.
- New flags: `--runtime`, `--node-version`, `--root-dir`, `--install-command`, `--build-command`, `--start-command`, and a repeatable `--static /=client/dist`. `--preset` and `--output-dir` are gone.
- `--prebuilt` keeps the old local-build path for Modelence apps.

## `modelence.json` and detection

`src/bin/appSpec.ts` defines the file the CLI reads and sends; Studio validates and merges it. The CLI never resolves the final plan itself — it sends three layers (file, flags, detected) and the server decides — but it prints a preview with per-field provenance, so each line says whether it came from the file, a flag, `engines`, the lockfile, or a profile.

Detection under `src/bin/detect/` is a pipeline rather than a pile of conditions: `facts.ts` does all the I/O once, then pure steps run in order — origin-blind **conventions** (lockfile → install, `engines.node` → version, scripts/Procfile → build and start, single startable workspace member → scoped build), then **frameworks** (a build with no start is a static site, served with SPA fallback), then at most one **profile** for where the project came from.

Two profiles exist today, and origin names appear nowhere else:

- `profiles/replit.ts` turns the artifact workspaces into static mounts in front of the API, builds each frontend with its `BASE_PATH`, and names `DATABASE_URL` and the Clerk keys up front.
- `profiles/lovable.ts` names the Supabase `VITE_*` variables.

Adding an origin is a new profile file plus a fixture.

`detect/commands.ts` centralizes package-manager differences, including the two that bit us in testing: pnpm is pinned from `packageManager` or the lockfile version instead of installing whatever is latest, and npm falls back to `npm install` when the lockfile is out of sync with `package.json` instead of failing `npm ci`.

## Test plan

- [x] Unit tests: seven golden fixtures with checked-in `expected.json` (express, vite site, modelence app, pnpm workspace, replit artifacts, lovable stale lock, yarn Procfile), so changing detection for one project kind fails that kind's test by name. Plus parsers, source packing and credential filtering, the auth cache, and spec merging.
- [x] `tsc` and eslint clean; `tsup` build succeeds.
- [ ] End-to-end against a dev Studio: `modelence init` then `modelence deploy` on a Replit-exported project (API + mounted frontend), a Lovable static site, and a Modelence app via `--prebuilt`.

> Requires the matching Studio changes (app spec resolution, runtime entrypoint, `/schema/modelence.json`) to be deployed first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
> Changes core deploy, Studio API integration, and on-disk auth caching—any bug could affect production releases, leak credentials in uploads, or break existing Modelence app deploys.
> 
> **Overview**
> **`modelence deploy`** is reworked so the default path uploads source and builds on Modelence Cloud from a root **`modelence.json`** contract, instead of always requiring `-a`/`-e` and a local `.modelence/build` bundle. Optional flags, a saved target in **`.modelence/project.json`**, or a browser deploy picker choose the environment; auth flows **`MODELENCE_TOKEN` → cached deploy token → browser**, with one re-auth on 401 without starting a second deployment. Legacy behavior remains for **`--prebuilt`** and for framework apps that depend on `modelence` but have no spec file.
> 
> New **`modelence init`** writes a **`modelence.json`** template (with schema URL) and prints the agent setup prompt; **`modelence logout`** revokes and clears cached logins. The CLI validates/parses JSONC specs, packs uploads via **`git ls-files`** when possible (with env/registry credential stripping and a size cap), streams deploy status/logs, and waits on provisioning before upload.
> 
> Docs add **`deploy/setup`** (agent-oriented reference and examples) plus a **Deploying any Node.js app** section in the AI agents guide; **`jsonc-parser`** is added as a dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bc632e1385577aaf52cf28493ff30b083d4543a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


artahian (commented):
Same notes as on the Studio PR: JSON config is fine for now, keep the env `type` in the schema and docs, but drop `type: 'modelence'` from resources. Configs and env vars stay separate concepts, but both should work for every app type.

eduard-piliposyan:
@omegascorp I separated one as a potential item that we can do afterward as a separate issue. But for other points, you can consider opening a separate issue and handling them afterward as well.

Scope: only things that can break and gaps in the design.

## Critical

- [ ] **`modelence deploy` hangs in CI or scripts instead of failing**

  `deploy.ts:86-127`, `auth.ts:79`. The CLI never checks whether a person is at the terminal. Whenever it has no token or no target, it opens the browser sign-in and polls for 10 minutes. Nobody will ever click in CI, a GitHub Action or an agent sandbox.
  - **Scenario 1:** `MODELENCE_TOKEN` is set, but no `-a/-e` is given and there is no `.modelence/project.json`. Framework templates gitignore `.modelence/`, so on a fresh CI checkout there never is one. The job hangs for 10 minutes and then fails.
  - **Scenario 2:** `MODELENCE_TOKEN` has expired. The 401 retry (`signInAgain`) opens the browser. This can also happen in the middle of `followDeploy`.
  - **Also:** tokens only come from the browser flow and last 1 hour, so there is no way to get a token CI can use.
  - **Fix:**
    - When `process.env.CI` is set or `!process.stdin.isTTY`, never open the browser. Fail at once with "pass `-a/-e` and `MODELENCE_TOKEN`".
    - Follow up with long-lived, revocable deploy keys scoped to an environment, created in the dashboard.

- [ ] **A one-off `-e production` becomes the default for every later deploy**

  `deployUpload.ts:83-89` rewrites `project.deploy` after **every** deploy, including ones with explicit flags. `deployTarget.ts:25` then uses the saved target when no flags are given.
  - **Scenario:** you run `modelence deploy -e production` once for a hotfix. Every later plain `modelence deploy` from that folder then goes to production, with no prompt. Before this PR `-a/-e` were required, so a deploy was always explicit.
  - **Fix:**
    - Save the default only when it was picked in the browser, not when it came from flags.
    - Print the target before building.
    - Ask for confirmation (or `--yes`) when the target differs from the saved default.

- [ ] **New CLI against the old Studio silently builds the wrong thing, so the release order must be fixed**

  The CLI now sends `{ environmentId, kind, spec }` (`deployUpload.ts:67-81`). The old Studio ignores `kind` and `spec`.
  - **Environments picked in the browser:** `environmentId` comes back `undefined`, and `/api/deploy` returns 400.
  - **Source deploys with `-a/-e`:** an Express or Vite app is built with the **prebuilt** Dockerfile (`npm install --omit=dev && npm start`, no build step). It "deploys", but it isn't what the file describes.
  - **Old CLI (≤0.25) against the new Studio:** checked, and it works.
  - **The two PRs give opposite release orders.** #385 says Studio goes first. Studio#484's `DEPLOYMENT.md` says #385 goes first.
  - **Fix:**
    - Agree on this order: `@modelence/runtime` publish → Studio deploy → docs → CLI npm publish.
    - Keep `0.26.0` on a `next`/`dev` tag until Studio is live.
    - Have the CLI stop with "Studio too old" when `/api/upload-bundle` returns no `environmentId`, instead of continuing with `undefined`.

- [ ] **Health check and port binding: deploys that fail or look healthy while down**

  Part of this is in this PR's `@modelence/runtime`, and part is in Studio's task definition (a TCP check on `localhost:3000`).
  - **Wait with no timeout:** `process.ts:70` `waitForApp` has no timeout and checks only `127.0.0.1`. In start + static mode, an app that listens on another port, or only on IPv6 or the task IP, leaves the router waiting forever, and the only signal is a failed deploy.
  - **Next.js standalone** (`output: 'standalone'`) listens on `$HOSTNAME`, which on Fargate is the task hostname, not `0.0.0.0`, so it fails. Next's own Dockerfile sets `HOSTNAME=0.0.0.0` for exactly this reason.
  - **Bound to `127.0.0.1` only:** the app passes a localhost TCP check but receives no traffic, so the deploy is reported successful while the site is down.
  - **Health = port open.** An app that opens its port before its DB is connected gets traffic too early, and a broken new version can replace healthy tasks.
  - **Fix:**
    - Set `HOSTNAME=0.0.0.0` in the image.
    - Give `waitForApp` a timeout, with an error that names the port.
    - Add an optional `healthCheck: { path, gracePeriodSeconds }` to the resource and use it as the rollout gate: an HTTP check against the task IP, not localhost.
    - Optionally add `port`, for apps that ignore `PORT`.

## High

- [ ] **The public contracts assume one service; fix the shape now, before the CLI ships**

  `resources` is a map, which is good, but everything around it is single-service. Once this CLI version is on npm, changing these shapes is a breaking change:
  - `ResolvedSpec` (`deploySpec.ts:84-89`) is `{ runtime, build, web }`, not per resource.
  - `/api/deploy/status` returns one `status`, `rolloutProgress`, `logs` and `siteUrl`.
  - The runtime reads one `MODELENCE_WEB`.
  - `env` is only top-level, not per resource.
  - There is no `type` on a resource (web / worker / cron / later `mongodb`, `redis`…). The key was removed in the last revision, and adding a required discriminator later breaks existing files.
  - **Fix, cheap now:** make the API responses keyed by resource name even with one entry. Reserve `type` (default `"web"`) and a per-resource `env` in the schema. The deeper backend work (ECS service per resource, internal networking, scaling) can come later.

- [ ] **`modelence.config.json` next to `modelence.config.ts` silently switches a framework app's deploy path**

  `deployKind.ts:40`: the presence of `modelence.config.json` alone moves a framework app from the local-build bundle path to a remote source build. Every framework app already has `modelence.config.ts`, a different file with the same stem and a totally different meaning.
  - **Scenario:** a user, or their agent, adds `modelence.config.json` thinking it is the JSON form of the config. The next deploy is built remotely on the `node` runtime with different commands and base image. Studio's git publish of the same app still pins the `modelence` runtime, so the same app is built two different ways.
  - **Fix:**
    - Pick a name that can't be confused: `modelence.json` (the earlier name) or `modelence.deploy.json`. The name freezes once published.
    - When a project depends on `modelence` and also has the file, print which path was chosen, or require `--source`.

- [ ] **What gets uploaded differs from what is in git, silently**

  `source.ts:151-163`:
  - **Symlinks** are dropped, because `lstat().isFile()` is false for them. They aren't even counted as excluded. This is common in monorepos and shared asset folders.
  - **Git submodules** are dropped: the entry is a directory, so it's skipped.
  - **Committed `.env*` files are removed** (`isLocalEnvFile`). Inside git, ignored `.env` files are already left out by `--exclude-standard`, so this rule only hits files the user *chose* to commit, such as `.env.production` with public `VITE_*` / `NEXT_PUBLIC_*` values. The cloud build then produces a different bundle than a local build.
  - **Fix:**
    - Keep symlinks that point inside the project (`archive.symlink`).
    - Fail or warn on submodules, or use `--recurse-submodules`.
    - Don't drop *tracked* `.env*` files in git mode; warn by name instead.
    - Print the excluded paths, not just a count.

- [ ] **No release / migration step, and the docs advise cron inside the web process**
  - **Migrations:** there is no `release` command, so migrations (Prisma, Drizzle, Knex…) can only go in `start`. They then run once per container, in parallel, on every restart and scale-out. **Fix:** add `resources.<name>.release`, run once with the new image before the rollout, failing the deploy if it fails. The key can be added later, but plan it into the deploy pipeline now.
  - **Cron advice:** `docs/web/deploy/setup.mdx` tells agents to move workers and cron loops into the web process (`setInterval`). With 2+ replicas, every job runs N times (emails, billing). Remove that advice, or add a warning that it needs a lock.

## Worth a quick fix (not blocking)

- [ ] **`proxy.ts:57-95`:** when the browser disconnects, the upstream request to the app is never aborted, so SSE and long-poll connections leak until the app closes them. Add `response.on('close', () => upstream.destroy())`.
- [ ] **Stale comments:** the PR description, `auth.ts` and `authCache.ts` disagree on token lifetime. Studio uses 1 hour. With the 40-minute safety margin, a cached login is reused for only 20 minutes, so users will see the browser almost every deploy.

---

## Separate issues to open (not blocking this PR)

- [ ] **App containers can't start while Studio is unreachable (not new)**
  - **Today (framework apps on prod):** `app/index.ts:153-162` awaits `connectCloudBackend()` → `POST /api/connect` (`app/backendApi.ts:65`) before anything else starts, with **no retry**. If it throws, the app doesn't start. The MongoDB URI arrives through those configs too. So during a Studio outage, any restart, scale-out or deploy of a tenant app fails today.
  - **With this PR:** `@modelence/runtime` (`env.ts`) adds `GET /api/env` for every container: 5 attempts over about 20s, and **no fetch timeout**, so a *hung* Studio stalls each attempt. Generic apps now get the same dependency. This isn't a regression, but it spreads the problem to every app type.
  - **Proposal:**
    - Short term: add a fetch timeout and keep retrying instead of exiting.
    - Long term: data-plane injection (ECS task `secrets` from SSM or Secrets Manager, written when deploy or restart runs), plus a last-known-good config cache for `/api/connect`.

eduard-piliposyan (changes requested):
Wrote issues in the other post

eduard-piliposyan (approved):
Only one comment, and approving.

**Docs: don't advise running migrations in `start.commands`**

Agents follow this page to the letter, so every generic app with a SQL database will end up with `prisma migrate deploy` (or Drizzle, Knex, TypeORM) as its first start command. Studio joins `start.commands` with `&&` (`resolve.ts:160`), so the migration runs in every container, before the port opens:

- **Concurrent migrations:** a rollout starts N new tasks at once, so N migrations run in parallel. Prisma takes a DB lock, but Drizzle, TypeORM and hand-written scripts usually don't.
- **Killed mid-migration:** a migration that runs longer than the health check's ~60s start period is killed partway through, and the task then crash-loops on a half-migrated schema.
- **Repeated runs:** it runs again on every restart and scale-out.

A proper `release` step (run once per deploy, before the rollout) can come later. For now, the guidance should keep agents from wiring this in by default:

```suggestion
   or, for a pure frontend, a static mount and no `start`. Do not add database migrations to
   `start.commands`: every container runs them on every start, several at once during a rollout,
   and a slow one is killed by the health check partway through. Leave migrations out of the file
   and tell the user to run them against the environment's database before deploying. A step may go
   first in `start.commands` only if it is fast and safe to run in parallel, such as `prisma generate`.
```

